PR #67, opened 2024-11-15
feat: support docker language

bxb100:
close: #49

## TODO

- [x] add test suits
- [x] test inner container
	- disable MacOS test, because it's without docker[^1]
	- disable Windows test, because it's without linux kernel[^2]
- [x] test windows

[^1]: https://github.com/actions/runner-images/issues/17
[^2]: https://github.com/orgs/community/discussions/21317


Cargo.toml
```diff
@@ -24,9 +24,11 @@ home = "0.5.9"
 indicatif = "0.17.8"
 indoc = "2.0.5"
 itertools = "0.13.0"
+md5 = "0.7.0"
 owo-colors = "4.1.0"
 rand = "0.8.5"
 rayon = "1.10.0"
+regex-lite = "0.1.6"
```
j178: Can we reuse the fancy-regex crate?

bxb100: Ok, I will fix it in the next commit

j178:
Thanks for your fantastic work! I'll review it soon!

bxb100:
@j178 I want to rewrite all commands to this crate https://github.com/fussybeaver/bollard , so maybe late review /= =/

j178:
@bxb100 I would prefer using the Docker CLI because it is lightweight and doesn't require many dependencies for basic Docker operations, unlike `bollard`, which seems a bit heavy to me. What do you think?

bxb100:
@j178 Ok

j178:
Thanks for your fantastic work here! I'm planning to make some adjustments to the coding style, hope you don't mind. Also, I'm considering simplifying the Docker test by removing the inner Docker test. It seems unnecessary to add a complex test job just for that.

j178:
The `typos-docker` image is around 500MB, it might not be suitable for a test case (It's too slow to run on my local machine).

bxb100:
I considered this and added a `nextest` profile inner `.config/nextest.toml` to prevent test docker(that does not work with `cargo test`).

maybe we can preserve it and add `#[ignored]`, only active this on github CI

bxb100:
I plan to modify `ci-docker.yml` to `ci-expensive.yml`, leverage ignore cfg and nextest filter to disable local test, but work on CI

j178:
(I'll handle the conflicts with master)

j178:
Thanks for your patience, let's merge!

PR #74, opened 2024-11-18
Add copyright of the original pre-commit to LICENSE

j178:
Thanks @asottile for pointing out that I actually violated the MIT license—my apologies for that!

Based on [this answer](https://opensource.stackexchange.com/a/5488), I think we can consolidate the original copyright statement into a single LICENSE file. I hope I'm doing it correctly.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,7 @@
 MIT License
 
 Copyright (c) 2024 j178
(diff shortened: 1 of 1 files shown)


asottile:
I would recommend keeping the entire license file for all of the derivative works (and an indication where they came from).  it's not just pre-commit that's plagiarized here but at least also identify, cfgv, and probably more.

PR #83, opened 2024-11-19
feat(sync): add auto sync fork action

lxb1226:
hi, this PR is to enable the auto-sync fork action.

### DIFF
--- a/.github/workflows/sync.yml
+++ b/.github/workflows/sync.yml
@@ -1,0 +1,40 @@
+name: Upstream Sync
+
+permissions:
+  contents: write
+
+on:
+  schedule:
+    - cron: "0 0 * * *" # every day
+  workflow_dispatch:
+
+jobs:
+  sync_latest_from_upstream:
+    name: Sync latest commits from upstream repo
+    runs-on: ubuntu-latest
+    if: ${{ github.event.repository.fork }}
+
+    steps:
+      # Step 1: run a standard checkout action
+      - name: Checkout target repo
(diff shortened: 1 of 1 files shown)


j178:
Sorry, could you clarify the purpose of this PR? I'm not sure this belongs in the repo.

lxb1226:
Sometime, when we fork a repository, we want to automatically sync the fork to keep it up-to-date with the upstream repository. like this [sync-a-fork](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/syncing-a-fork) , but it is automatically.

j178:
Shouldn't this action be in the fork instead?

lxb1226:
Yes, this action will be enable in the fork. However, we need to stay consistent with the upstream, so this YAML file needs to be in the upstream repository.

j178:
Thanks, but I really don't think it's reasonable to put an action like this in the upstream repo just for some forks to auto sync. Not every fork wants to be auto synced I think, but this action cannot be opt-out.

lxb1226:
This might be a bit misleading, as even in the fork, this action needs to be manually enabled, so it’s only useful for those who need it.

j178:
Yeah, forks might enable action because they want to run the CI job, but this cron job auto runs every day unless them turn off the entire action.

I think a better solution is to place this commit at the top of your fork master branch. Let the sync action rebase during each sync so that this commit remains exclusive to the fork.

Issue #87, opened 2024-11-21
dynamic run expensive CI


bxb100:
Currently, we have some expensive CI checks, eg. docker. 

I want to add a dynamic run workflow like [OpenDAL](https://github.com/apache/opendal/blob/main/.github/scripts/test_behavior/plan.py).

```mermaid
stateDiagram-v2
    [*] --> Push
    [*] --> PR

	State1: Language CI
    note left of State1
      every expensive language CI can be set differently
    end note

    Push --> Determine
    PR --> Determine
    Determine --> State1
    State1 --> [*]
```

j178:
Thanks, looks reasonable to me!

PR #99, opened 2024-11-22
ci: dynamic run expensive test

bxb100:
close: #87

- remove `docker` feature
- expensive test task set in `.github/actions/(macos|windwos|ubuntu)/{lang}/action.yml`
- determine modified files using https://github.com/dorny/paths-filter

---

example: https://github.com/bxb100/pre-commit-rs/actions/workflows/ci-expensive.yml?query=branch%3ACI-demo2

test-branch: `CI-demo` `CI-demo2`

### DIFF
--- a/.github/actions/dynamic/action.yml
+++ b/.github/actions/dynamic/action.yml
@@ -1,0 +1,41 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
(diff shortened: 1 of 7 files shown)


j178:
Thanks, gonna review soon!

j178:
This is a very clever approach. However, I'm hesitant to merge it right away, as our current Docker test isn't very costly and it's the only test that would benefit from this PR. I greatly appreciate your effort to improve this. In the future, if we have more expensive tests, I will definitely advocate for this change.

bxb100:
ok

PR #101, opened 2024-11-22
feat: support init-templatedir

### LINKED ISSUE #32: Implement `pre-commit init-templatedir`

XmchxUp:
Closed #32

### DIFF
--- a/src/cli/mod.rs
+++ b/src/cli/mod.rs
@@ -16,7 +16,7 @@ mod validate;
 
 pub(crate) use clean::clean;
 pub(crate) use hook_impl::hook_impl;
-pub(crate) use install::{install, uninstall};
+pub(crate) use install::{init_template_dir, install, uninstall};
 pub(crate) use run::run;
 pub(crate) use sample_config::sample_config;
 pub(crate) use self_update::self_update;
--- a/src/cli/mod.rs
+++ b/src/cli/mod.rs
@@ -343,13 +343,13 @@ pub(crate) struct GenerateShellCompletionArgs {
 #[derive(Debug, Args)]
 pub(crate) struct InitTemplateDirArgs {
     /// The directory in which to write the hook script.
-    pub(crate) directory: String,
+    pub(crate) directory: PathBuf,
 
     /// Assume cloned repos should have a `pre-commit` config.
-    #[arg(long, default_value_t = false)]
+    #[arg(long)]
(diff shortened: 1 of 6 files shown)


src/cli/mod.rs
```diff
@@ -339,3 +339,17 @@ pub(crate) struct GenerateShellCompletionArgs {
     #[arg(value_enum)]
     pub shell: clap_complete::Shell,
 }
+
+#[derive(Debug, Args)]
+pub(crate) struct InitTemplateDirArgs {
+    /// The directory in which to write the hook script.
+    pub(crate) directory: String,
```
j178:
```suggestion
    pub(crate) directory: PathBuf,
```

src/cli/mod.rs
```diff
@@ -339,3 +339,17 @@ pub(crate) struct GenerateShellCompletionArgs {
...
+
+    /// Assume cloned repos should have a `pre-commit` config.
+    #[arg(long, default_value_t = false)]
```
j178:
```suggestion
    #[arg(long)]
```

XmchxUp: I think this is an option, not an argument.

src/cli/mod.rs
```diff
@@ -339,3 +339,17 @@ pub(crate) struct GenerateShellCompletionArgs {
...
+    pub(crate) no_allow_missing_config: bool,
+
+    /// Which hook type to install.
+    #[arg(long, short = 't', default_value_t = HookType::PreCommit)]
+    pub(crate) hook_type: HookType,
```
j178:
```suggestion
    #[arg(short = 't', long = "hook-type", value_name = "HOOK_TYPE", value_enum)]
    pub(crate) hook_types: Vec<HookType>,
```

j178:
Thanks! Could you add a test case too?

XmchxUp:
> Thanks! Could you add a test case too?

Thanks for the review, will add it later.

XmchxUp:
@j178 

Should `init-templatedir`, which is based on `install`, use a mock library to simulate the behavior of install for testing, or is there another method?

j178:
We don’t need to use mocks in integration tests. You can refer to other tests in the `tests` folder for guidance.

j178:
I made some tweaks and added a test, thanks for your contribution!

PR #113, opened 2024-11-24
feat: support docker_image

bxb100:
close: #50


src/languages/docker.rs
```diff
@@ -212,6 +212,7 @@ impl LanguageImpl for Docker {
                     .args(&cmds[1..])
                     .args(hook_args.as_ref())
                     .args(batch)
+                    .check(false)
```
bxb100, line 215: Hi, @j178 I set this because I think the user needs to know the container failed info

j178:
Thanks!

PR #117, opened 2024-11-25
refactor: revert #107 but using `enum_dispatch`

bxb100:
hi, @j178 , IMHO I think using [enum_dispatch](https://docs.rs/enum_dispatch/latest/enum_dispatch/) to remove boilerplate code much better, PTAL

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -489,6 +489,18 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "a357d28ed41a50f9c765dbfe56cbc04a64e53e5fc58ba79fbc34c10ef3df831f"
 
 [[package]]
+name = "enum_dispatch"
+version = "0.3.13"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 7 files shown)


j178:
Thanks for letting me know about `enum_dispatch`. Before #107 (and in this PR), we had to perform two matches: first from `config::Language` to `languages::Language`, and then from `languages::Language.<some method>` to the actual function implementation. After #107, we only need to perform one match.

There is some boilerplate code, but it seems manageable. We could write a macro to reduce the boilerplate code into a single location. What are your thoughts on this?

bxb100:
I prefer not to use complicated macros since they are often difficult to read and understand. However, I am open to it, if you have some good ideas, please let me know

bxb100:
understand, my initial motivation was to eliminate `static`, if you want to keep this, I'm fine with it

bxb100:
Hi, I’m temporarily closing this. If you need it in the future, I’ll be happy to reopen it.

j178:
Thanks!

PR #181, opened 2025-02-17
Restructure folders layout

j178:
Restucture the `~/.cache/prefligit/` directory layout, moving from the `pre-commit` implementation.

- No longer use `sqlite` db
- Move cloned repositories into `HOME/repos`
- No longer create dummy packages for local repos
- Move installed hooks environment into `HOME/hooks`

### DIFF
--- a/src/hook.rs
+++ b/src/hook.rs
@@ -1,5 +1,6 @@
 use std::collections::{HashMap, HashSet};
 use std::fmt::Display;
+use std::hash::Hash;
 use std::path::{Path, PathBuf};
 use std::rc::Rc;
 use std::sync::Mutex;
@@ -37,7 +38,7 @@ pub enum Error {
 #[derive(Debug, Clone)]
 pub enum Repo {
     Remote {
-        /// Path to the stored repo.
+        /// Path to the cloned repo.
         path: PathBuf,
         url: Url,
         rev: String,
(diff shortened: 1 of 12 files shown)


src/hook.rs
```diff
@@ -489,67 +458,80 @@ pub struct Hook {
...
+
+impl Hash for Hook {
+    fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
+        match &*self.repo {
+            Repo::Remote { url, rev, .. } => {
+                url.hash(state);
+                rev.hash(state);
+            }
+            Repo::Local { .. } => {
+                "local".hash(state);
```
j178, line 528: use ToString

src/languages/docker.rs
```diff
@@ -23,13 +23,17 @@ pub struct Docker;
...
-            .and_then(OsStr::to_str)
-            .map(|s| format!("pre-commit-{:x}", md5::Md5::digest(s)))
+        let path = hook.env_path()?;
+        let mut hasher = SeaHasher::new();
+        path.hash(&mut hasher);
+        Some(format!("pre-commit-{:x}", hasher.finish()))
     }
 
     async fn build_docker_image(hook: &Hook, pull: bool) -> Result<()> {
+        let Some(src) = hook.repo_path() else {
```
j178, line 33: warning if additional deps not empty

src/languages/docker.rs
```diff
@@ -162,12 +166,14 @@ impl Docker {
...
 impl LanguageImpl for Docker {
-    fn environment_dir(&self) -> Option<&str> {
-        Some("docker")
+    fn supports_dependency(&self) -> bool {
+        true
     }
 
     async fn install(&self, hook: &Hook) -> Result<()> {
-        let env = hook.environment_dir().expect("No environment dir found");
+        let Some(env) = hook.env_path() else {
```
j178, line 174: make a type to guarantee this is always Somw

src/store.rs
```diff
@@ -25,8 +27,6 @@ pub enum Error {
     #[error(transparent)]
     Fmt(#[from] std::fmt::Error),
     #[error(transparent)]
-    DB(#[from] rusqlite::Error),
```
j178, line 28: remove LocalHookNoEnv

PR #251, opened 2025-05-01
Add language name to 'not yet implemented' messages

qrichert:
Hi, this commit makes the 'not yet implemented' message a bit more meaningful by adding the name of the language that's not yet supported.

## Before

before

## After

after


j178:
Thanks!

PR #326, opened 2025-08-03
Improve config parsing error message

j178:


j178:
I don't like the approach here, hoping someone teach me how to do this better :D

PR #332, opened 2025-08-03
Fix large file check to use staged_get instead of intent_add

lemorage:
### Reason
`intent_to_add_files()` returns files that are only marked for tracking, but have no staged contents. Since there is no actual content to size-check, it would give false positives.

### Solution
Use `get_staged_files()` instead.

### DIFF
--- a/tests/common/mod.rs
+++ b/tests/common/mod.rs
@@ -127,6 +127,12 @@ impl TestContext {
         command
     }
 
+    pub fn run_all_files(&self) -> Command {
+        let mut command = self.run();
+        command.arg("--all-files");
+        command
+    }
+
     pub fn clean(&self) -> Command {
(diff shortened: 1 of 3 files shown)


j178:
Thanks!
Since you are here, would you mind adding a test for this too?

lemorage:
Sure!

tests/common/mod.rs
```diff
@@ -127,6 +127,12 @@ impl TestContext {
         command
     }
 
+    pub fn run_all_files(&self) -> Command {
+        let mut command = self.run();
+        command.arg("--all-files");
+        command
+    }
+
```
j178: We usually add dedicated function only for the top-level commands, this function isn't necessary I think.

j178:
Thank you!

PR #341, opened 2025-08-04
Use DFS to find connected components in hook dependencies

### LINKED ISSUE #336: Improve hook installs concurrency
Hooks that dependencies has no overlappings can be installed in parallel.

j178:
Closes #336


j178:
image

PR #350, opened 2025-08-05
Fix bug that directories are ignored

j178:
Also add a friendly warning for files that aren't exist.

And this PR moves files exist check to only in `--files` case, this should improve performance for `--all-files`.


j178:
Relates to #317

PR #352, opened 2025-08-05
Check minimum supported version for uv found from system

### LINKED ISSUE #295: Check minimum supported version for uv found from system

lemorage:
Close #295

### DIFF
--- a/src/languages/python/uv.rs
+++ b/src/languages/python/uv.rs
@@ -3,7 +3,7 @@ use std::path::{Path, PathBuf};
 use std::sync::LazyLock;
 use std::time::Duration;
 
-use anyhow::Result;
+use anyhow::{Result, bail};
 use axoupdater::{AxoUpdater, ReleaseSource, ReleaseSourceType, UpdateRequest};
 use semver::Version;
 use tokio::task::JoinSet;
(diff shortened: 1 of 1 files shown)


src/languages/python/uv.rs
```diff
@@ -221,32 +222,64 @@ impl Uv {
...
+
     pub async fn install(store: &Store) -> Result<Self> {
-        // 1) Check if `uv` is installed already.
-        // TODO: check minimum supported uv version
-        if let Ok(uv) = UV_EXE.as_ref() {
-            return Ok(Self::new(uv.clone()));
+        // 1) Check if system `uv` meets minimum version requirement
+        if let Ok(uv_path) = UV_EXE.as_ref() {
+            if let Ok(version) = Self::get_uv_version(uv_path).await {
+                let min_version = Version::parse(UV_VERSION)?;
```
j178: I’m thinking of splitting UV_VERSION into UV_VERSION_MIN and UV_VERSION_MAX. The idea is to set a lower and upper limit.

When we check the system for a uv version, we’d use both UV_VERSION_MIN and UV_VERSION_MAX to ensure we don’t pick a version that’s either too old and missing features or too new and potentially has some breaking changes. And when installing a new uv version, we’d use UV_VERSION_MAX to aim for something close to the latest.

Since uv is pretty stable, we can have a pretty broad range between MIN and MAX. That way, we won’t need to install a managed uv as often.

What do you think?

lemorage: Good idea! What value do u suggest for setting MIN and MAX?

j178: We can bump MAX to `0.8.4`, which is the latest version. 
For MIN, I’m leaning towards `0.7.0` to support the last two minor releases. What do you think?

src/languages/python/uv.rs
```diff
@@ -221,32 +222,64 @@ impl Uv {
...
+    }
+
     pub async fn install(store: &Store) -> Result<Self> {
-        // 1) Check if `uv` is installed already.
-        // TODO: check minimum supported uv version
-        if let Ok(uv) = UV_EXE.as_ref() {
-            return Ok(Self::new(uv.clone()));
+        // 1) Check if system `uv` meets minimum version requirement
+        if let Ok(uv_path) = UV_EXE.as_ref() {
+            if let Ok(version) = Self::get_uv_version(uv_path).await {
```
j178: I also want to move the quering version logic into the `UV_EXE` LazyLock, so we don't need to query multiple times. And change from `which::which("uv")` to `which::which_all("uv")` to find all system uv.

src/languages/python/uv.rs
lemorage: All good!

src/languages/python/uv.rs
```diff
@@ -14,13 +16,49 @@ use crate::fs::LockedFile;
...
+    for uv_path in which::which_all("uv").ok()? {
+        debug!("Found uv in PATH: {}", uv_path.display());
+
+        if let Ok(version) = get_uv_version(&uv_path) {
+            if max_version >= version && version >= min_version {
+                return Some((uv_path, version));
+            }
+            warn!(
+                "System uv version {} is not in the acceptable range between {} and {}, \
+                     pre-commit-hooks will install its own version.",
```
j178:
```suggestion
```

j178: I'd prefer just deleting this line.

j178:
Thank you, this is very nice!

PR #367, opened 2025-08-06
Fix bug that default hook stage should be pre-commit

### LINKED ISSUE #366: `stages: ["manual"]` is run via `prefligit run -a`
Apache Airflow has a number of hooks marked with the "manual" stage in our .pre-commit-config.yaml that should only be run when the stage is explicitly requested, as per https://pre-commit.com/#confining-hooks-to-run-at-certain-stages.

j178:
Closes #366

### DIFF
--- a/tests/run.rs
+++ b/tests/run.rs
@@ -302,7 +302,7 @@ fn invalid_hook_id() {
     ----- stdout -----
 
     ----- stderr -----
-    No hook found for id `invalid-hook-id`
+    No hook found for id `invalid-hook-id` and stage `pre-commit`
     "#);
 }
 


potiuk:
nice :)

ashb:
@j178 A nit I've noticed on this one: `pre-commit run my-hook-with-stage-manual` will always work, but after this change I have to manually specify `prek run --hook-stage manual my-hook-with-stage-manual`

j178:
@ashb  Interesting! I tried running `pre-commit run my-hook-with-stage-manual` and it didn’t work for me. I checked out the pre-commit source code here:

https://github.com/pre-commit/pre-commit/blob/72a3b71f0e7b89b78899739aeee2e962914ab649/pre_commit/commands/run.py#L423-L428

It looks like it’s filtering hooks by both hook-id and stage. Could you double-check that part?

ashb:
.... You are correct. I could have _sworn_ that worked before. I must be an AI (i.e. hallucinating)

PR #378, opened 2025-08-07
Fix identifying symlinks

branchvincent:
Hey, cool project!

When trying to replace `pre-commit` in my repo, I noticed symlinks weren't being identified since `std::fs::metadata` [resolves symlinks](https://doc.rust-lang.org/std/fs/fn.metadata.html):

> This function will traverse symbolic links to query information about the destination file.


j178:
Thanks! Should we use https://doc.rust-lang.org/std/fs/fn.symlink_metadata.html instead of std::fs::metadata?

branchv:
Yea makes sense, done

PR #380, opened 2025-08-07
Add dynamic completion of hook ids

### LINKED ISSUE #370: Implement hook name autocomplete
In Airflow we have 170+ pre-commits. We have our own wrapper to pre-commit `breeze static-checks` and the main reason we have the wrapper (we have few others but less important) is that we wanted to have auto-complete working for our pre-commits, we use pretty consistent naming scheme followin "update-...", "check-..." , "generate-..." and it saves a ton of time and focus if you can just do:

```
breeze static-checks --type check-provider<TAB>` and see and autocomplete the right check from the list.
```

It would be awesome if we do not have to wrap it in our breeze tool, but use:

```
prefligit run check-provider<TAB>
```

We tried to discuss adding such feature to pre-commit, but the pre-commit author @asotile had quite unreasonable expectations and heavily insisted (including marking a number of valid comments from people as off-topic) that the latency shoudl be extremely low and demanded to write in-memory server that would be automatically started when <TAB> is entereed for better performance. Even after explaining that simple caching available commands in local storage would be far more than enough due to in-memory caching of POSIX filesystems  - I was not able to convince the pre-commit author that we could do it this way (see https://github.com/pre-commit/pre-commit/issues/1119)

With prefligit soon replacing pre-commit for us (and likely many others) it would be great to have such functionality built in so that we do not have to add a wrapper around it - this way we could remove `breeze static-checks` eventually from Airflow's breeze development tool and simply use prefligit.

If that feature is welcome I might maybe even attempt to add it (though I'd have to learn rust a bit for that- but it's so close to my heart and so useful feature for thousands of Airflow contributors that I am willing to spend quite some time on it.

Would love to learn @j178 what are your thoughts on potential implementation here - maybe in rust reading and parsing the pre-commit config xml file is fast enough so that we do not even have to worry about caching and extracting available pre-commit names to a separate cache file?

Or maybe you also think that you need to start a separate server to cache pre-commit names in memory. I considered it pretty unreasonable approach, but maybe I am wrong and that's the only good way of doing it?

branchvincent:
Fixes #370

Just playing around with clap's unstable dynamic completions to see what's possible: https://github.com/clap-rs/clap/issues/3166. IIUC, it seems like the dynamic completion engine requires generating completions via [`CompleteEnv`](https://docs.rs/clap_complete/latest/clap_complete/env/struct.CompleteEnv.html) but maybe there's a way to keep the current `generate-shell-completion` command

Demo using fish:

```console
$ complete --erase prefligit; COMPLETE=fish cargo run | source
$ time complete -C './target/debug/prefligit run c'
cargo-fmt
cargo-clippy

________________________________________________________
Executed in   25.92 millis    fish           external
   usr time   17.36 millis   13.97 millis    3.38 millis
   sys time    8.14 millis    0.67 millis    7.47 millis
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -398,6 +398,9 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "67e4efcbb5da11a92e8a609233aa1e8a7d91e38de0be865f016d14700d45a7fd"
 dependencies = [
  "clap",
+ "clap_lex",
+ "is_executable",
+ "shlex",
 ]
 
 [[package]]
@@ -1249,6 +1252,15 @@ name = "is_ci"
 version = "1.2.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7655c9839580ee829dfacba1d1278c2b7883e50a277ff7541299489d6bdfdc45"
+
+[[package]]
+name = "is_executable"
+version = "1.0.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d4a1b5bad6f9072935961dfbf1cced2f3d129963d091b6f69f007fe04e758ae2"
+dependencies = [
+ "winapi",
+]
 
 [[package]]
 name = "is_terminal_polyfill"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,7 +36,7 @@ async_zip = { git = "https://github.com/astral-sh/rs-async-zip", rev = "c909fda6
 axoupdater = { version = "0.9.0", default-features = false, features = [ "github_releases"] }
 bstr = { version = "1.11.0" }
 clap = { version = "4.5.16", features = ["derive", "env", "string", "wrap_help"] }
-clap_complete = { version = "4.5.37" }
+clap_complete = { version = "4.5.37", features = ["unstable-dynamic"]}
 ctrlc = { version = "3.4.5" }
 dunce = { version = "1.0.5" }
 etcetera = { version = "0.10.0" }
--- a/src/cli/mod.rs
+++ b/src/cli/mod.rs
@@ -5,10 +5,12 @@ use std::process::ExitCode;
 use clap::builder::Styles;
 use clap::builder::styling::{AnsiColor, Effects};
 use clap::{ArgAction, Args, Parser, Subcommand, ValueHint};
+use clap_complete::engine::{ArgValueCompleter, CompletionCandidate};
 
 use constants::env_vars::EnvVars;
 
-use crate::config::{CONFIG_FILE, HookType, Stage};
+use crate::config::{self, CONFIG_FILE, HookType, Stage};
+use crate::workspace::Project;
 
 mod clean;
 mod hook_impl;
@@ -26,6 +28,39 @@ pub(crate) use run::run;
 pub(crate) use sample_config::sample_config;
 pub(crate) use self_update::self_update;
 pub(crate) use validate::{validate_configs, validate_manifest};
+
+// Parses hook ids from .pre-commit-config.yaml
+fn hook_id_completer(current: &std::ffi::OsStr) -> Vec<CompletionCandidate> {
+    let Ok(project) = Project::from_config_file(None) else {
+        return vec![];
+    };
+
+    let hook_ids = project
+        .config()
+        .repos
+        .iter()
+        .flat_map(|repo| -> Box<dyn Iterator<Item = &String>> {
+            match repo {
+                config::Repo::Remote(cfg) => Box::new(cfg.hooks.iter().map(|h| &h.id)),
+                config::Repo::Local(cfg) => Box::new(cfg.hooks.iter().map(|h| &h.id)),
+                config::Repo::Meta(cfg) => Box::new(cfg.hooks.iter().map(|h| &h.0.id)),
+            }
+        })
+        .map(|id| CompletionCandidate::new(id.clone()));
+
+    let Some(current) = current.to_str() else {
+        return hook_ids.collect();
+    };
+
+    hook_ids
+        .filter(|h| {
+            h.get_value()
+                .to_str()
+                .unwrap_or_default()
+                .starts_with(current)
+        })
+        .collect()
+}
 
 #[derive(Copy, Clone)]
 pub(crate) enum ExitStatus {
@@ -257,7 +292,7 @@ pub(crate) struct RunExtraArgs {
 #[derive(Debug, Clone, Default, Args)]
 pub(crate) struct RunArgs {
     /// The hook ID to run.
-    #[arg(value_name = "HOOK", value_hint = ValueHint::Other)]
+    #[arg(value_name = "HOOK", value_hint = ValueHint::Other, add = ArgValueCompleter::new(hook_id_completer))]
     pub(crate) hook_id: Option<String>,
     /// Run on all files in the repo.
     #[arg(short, long, conflicts_with_all = ["files", "from_ref", "to_ref"])]
--- a/src/cli/mod.rs
+++ b/src/cli/mod.rs
@@ -2,8 +2,8 @@ use std::ffi::OsString;
 use std::path::PathBuf;
 use std::process::ExitCode;
 
-use clap::builder::Styles;
 use clap::builder::styling::{AnsiColor, Effects};
+use clap::builder::{StyledStr, Styles};
 use clap::{ArgAction, Args, Parser, Subcommand, ValueHint};
 use clap_complete::engine::{ArgValueCompleter, CompletionCandidate};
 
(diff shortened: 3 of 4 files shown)


branchv:
cc @potiuk if you want to try this out

potiuk:
> cc @potiuk if you want to try this out

wowowowow

ashb:
zsh version:
```
❯ cargo install --git https://github.com/branchvincent/prefligit --branch hook-completions
❯ eval "$(COMPLETE=zsh prefligit)"
```

And then `preflight run<TAB>` gives this on Airflow:

```
❯ prefligit run                                                                                                                                                                              [10:01] [130] !27938
---- values
--all-files                                                -- Run on all files in the repo
--color                                                    -- Whether to use color in output
--config                                                   -- Path to alternate config file
--directory                                                -- Run hooks on all files in the specified directories
--files                                                    -- Specific filenames to run hooks on
--from-ref                                                 -- The original ref in a `from_ref...to_ref` diff expression. Files changed in this diff will be run through the hooks
--help                                                     -- Display the concise help for this command
--hook-stage                                               -- The stage during which the hook is fired
--last-commit                                              -- Run hooks against the last commit (HEAD~1..HEAD)
--no-progress                                              -- Hide all progress outputs
--quiet                                                    -- Do not print any output
--show-diff-on-failure                                     -- When hooks fail, run `git diff` directly afterward
--to-ref                                                   -- The destination ref in a `from_ref...to_ref` diff expression. Files changed in this diff will be run through the hooks
--verbose                                                  -- Use verbose output
--version                                                  -- Display the prefligit version
bandit                                                     check-pre-commit-information-consistent                    lint-helm-chart
blacken-docs                                               check-provide-create-sessions-imports                      lint-json-schema
check-aiobotocore-optional                                 check-provider-docs-valid                                  lint-markdown
check-airflow-k8s-not-used                                 check-provider-yaml-valid                                  mixed-line-ending
check-airflow-providers-bug-report-template                check-providers-subpackages-init-file-exist                mypy-airflow-core
check-apache-license-rat                                   check-pydevd-left-in-code                                  mypy-airflow-ctl
check-base-operator-partial-arguments                      check-revision-heads-map                                   mypy-dev
check-base-operator-usage                                  check-safe-filter-usage-in-html                            mypy-devel-common
check-boring-cyborg-configuration                          check-significant-newsfragments-are-valid                  mypy-providers
check-breeze-top-dependencies-limited                      check-sql-dependency-common-data-structure                 mypy-task-sdk
... [truncated by me for this issue] ...
```

Lovely!

One nit, that I'm not sure if it's possible to do is that a simple complete on `preflight` lists soo many options, so I wonder if we can be smarter about it somehow? I don't really know how mind you, just musing.

ashb:
Oh, we should probably use show the description alongside the completion, i.e. so given this

```yaml
      - id: pylint
        name: pylint
        description: "Pylint is a static code analyser for Python 2 or 3."
      - id: check-fab-migrations
        language: pygrep
        name: Check no migration is done on FAB related table
```

We'd see something like this:

```
--verbose                                                  -- Use verbose output
--version                                                  -- Display the prefligit version
pylint                                                     -- Pylint is a static code analyser for Python 2 or 3.
check-fab-migrations                                       -- Check no migration is done on FAB related table
```


Though I just noticed/rememberd that pre-commit allows multiple hooks with the same name, so I'm not sure how we should handle cases like this
https://github.com/apache/airflow/blob/5db2cf735ab49cf0bfd48547c1f360c716976e9e/.pre-commit-config.yaml#L87-L98


```yaml
      - id: insert-license
        name: Add license for all Shell files
        # ...
      - id: insert-license
        name: Add license for all toml files
        # ...
```

potiuk:
> Oh, we should probably use show the description alongside the completion, i.e. so given this

That would be FANTASTIC :)

potiuk:
> Though I just noticed/rememberd that pre-commit allows multiple hooks with the same name, so I'm not sure how we should handle cases like this

We could combine those - long but there are just a few.

potiuk:
Works nicely for me too. And it's FAST :D

j178:
Now it looks like this:
```console
❯ prefligit run -vvv <TAB>
identity                                                                                                                (identity)
check-hooks-apply                                                                                              (Check hooks apply)
doctoc                                                                                        (Add TOC for Markdown and RST files)
insert-license                                                                                     (Add license for all SQL files)
check-min-python-version                                                                            (Check minimum Python version)
update-black-version                                                                   (Update black versions everywhere (manual))
update-installers-and-pre-commit                                             (Update installers and pre-commit to latest (manual))
update-chart-dependencies                                                           (Update chart dependencies to latest (manual))
check-taskinstance-tis-attrs                                                      (Check that TI and TIS have the same attributes)
check-deferrable-default                                                       (Check and fix default value of default_deferrable)
blacken-docs                                                                                                   (Run black on docs)
check-merge-conflict                                                          (Check that merge conflicts are not being committed)
debug-statements                                                                  (Detect accidentally committed debug statements)
check-builtin-literals                                                         (Require literal syntax when initializing builtins)
…and 145 more rows
```

j178:
And I changed the match rule from `starts_with` to `contains`, I think if would be a little more friendly?

potiuk:
> Now it looks like this:

Fantastic. Works with `zsh` beautifully as well:

It only shows "last" pre-commit help, but it's "good enough":

Screenshot 2025-08-07 at 12 24 06

BTW. The `generate-shell-completion` command is not shown in `prefligit --help`  @j178  yet (I guess it's experimental)  - but the `eval "$(COMPLETE=zsh prefligit)"` works very well, so maybe it just needs documentation or a command to automatically add `eval` to the right `.*rc` files automatically ? We did it in `breeze` wrapper

Our `breeze setup autocomplete` checks which SHELL is used and checks if the autocomplete has been already added and adds it (we use python's click autcomplete generator for that so we have scripts in our repo that we can source). 

```
# START: Added by Updated Airflow Breeze autocomplete setup
source /Users/jarekpotiuk/code/airflow/dev/breeze/autocomplete/breeze-complete-zsh.sh
# END: Added by Updated Airflow Breeze autocomplete setup
```

potiuk:
> And I changed the match rule from starts_with to contains, I think if would be a little more friendly?

I have not see it working:

Screenshot 2025-08-07 at 12 30 41

We have more pre-commits with `provider` for example and they are not shown (just updated latest version)

potiuk:
> BTW. The generate-shell-completion command is not shown in prefligit --help @j178 yet (I guess it's experimental) - but the eval "$(COMPLETE=zsh prefligit)" works very well, so maybe it just needs documentation or a command to automatically add eval to the right .*rc files automatically ? We did it in breeze wrapper

that was comment to @j178 of course - to consider how to make autocomplete setup :)

j178:
> > And I changed the match rule from starts_with to contains, I think if would be a little more friendly?
> 
> I have not see it working:
> 
> Screenshot 2025-08-07 at 12 30 41
> We have more pre-commits with `provider` for example and they are not shown (just updated latest version)

I guess there is some filtering logic in the shell completion mechanism. If there are prefix matched candiates, then it just ignores all other substring matches.

```console
❯ prefligit run pr
pretty-format-json                                         (Format JSON files)
prevent-deprecated-sqlalchemy-usage      (Prevent deprecated sqlalchemy usage)
provider-version-compat  (Check for correct version_compat imports in provid…)
```

Only if there is no prefix match, it will show the candidates with substring match:

```console
❯ prefligit run ve
check-min-python-version                        (Check minimum Python version)
update-black-version               (Update black versions everywhere (manual))
check-executables-have-shebangs          (Check that executables have shebang)
prevent-deprecated-sqlalchemy-usage      (Prevent deprecated sqlalchemy usage)
update-supported-versions        (Updates supported versions in documentation)
update-version                                       (Update versions in docs)
check-links-to-example-dags-do-not-use-hardcoded…  (Verify no hard-coded ver…)
check-for-inclusive-language  (Check for language that we do not accept as c…)
update-vendored-in-k8s-json-schema  (Vendor k8s definitions into values.sche…)
mypy-devel-common                                  (Run mypy for devel-common)
provider-version-compat  (Check for correct version_compat imports in provid…)
check-airflow-version-checks-in-core  (No AIRFLOW_V_* imports in airflow-core)
```

potiuk:
> Only if there is no prefix match, it will show the candidates with substring match:

Then it's cool and likely the best of both worlds how it is now :)

potiuk:
BTW. It does not work with `ve` in `zsh` either :). But that's not an issue

potiuk (approved):

j178:
> BTW. The generate-shell-completion command is not shown in prefligit --help @j178 yet (I guess it's experimental) - but the eval "$(COMPLETE=zsh prefligit)" works very well, so maybe it just needs documentation or a command to automatically add eval to the right .*rc files automatically ? We did it in breeze wrapper

We hide it because it’s mainly used by package managers, like [Homebrew](https://github.com/Homebrew/homebrew-core/blob/5474481a5618a9c363d707765b8e34bb37488eed/Formula/p/prefligit.rb#L23). When you install something using a package manager, the completion files are automatically placed where they need to go to work properly.

I tried to replace current `generate-shell-completion` with something like:
```rust
        Command::GenerateShellCompletion(args) => {
            let command = Cli::command();


            let bin_name = command
                .get_bin_name()
                .unwrap_or_else(|| command.get_name())
                .to_owned();

            unsafe {
                std::env::set_var(EnvVars::COMPLETE, args.shell.to_string());
            }
            CompleteEnv::with_factory(Cli::command)
                .bin(bin_name)
                .var(EnvVars::COMPLETE)
                .complete();

            Ok(ExitStatus::Success)
        }
```

Right now, though, `CompleteEnv` generates completions with a hard-coded absolute path for the binary, which is a problem. 

```console
❯ COMPLETE=fish cargo run
complete --keep-order --exclusive --command prefligit --arguments "(COMPLETE=fish "'/Users/Jo/code/rust/prefligit/target/debug/prefligit'" -- (commandline --current-process --tokenize --cut-at-cursor) (commandline --current-token))"
```

It doesn’t let package managers like Homebrew create those completions while building and pass them on to the user, since the binary’s path might differ depending on the user’s setup.

potiuk:
> It doesn’t let package managers like Homebrew create those completions while building and pass them on to the user, since the binary’s path might differ depending on the user’s setup.

:(

potiuk:
That's still fine if we can have just documentation on how you can get the completion running IF you manually configure it in your .zshrc ... We can even add documentation about it to Airflow as a "prerequisite" if you want hook completion

j178:
Let’s wait and see what @branchvincent  thinks about this. Do you think it’s good for us to ship it as is and find a better way to integrate it with `generate-shell-completion` later?

potiuk:
> or us to ship it as is and find a better way to integrate it with `generate-shell-completion` later?

I personally do - maybe a few line doc about how to set it up for different shells to have the dynamic auto-completion work would need to be added, but other than that **it works for me**

branchv:
Thanks for adding the descriptions and other tweaks @j178, looks great!
 
> CompleteEnv generates completions with a hard-coded absolute path for the binary, which is a problem

It depends on how you call it

```console
$ PATH=$(pwd)/target/debug COMPLETE=fish prefligit
complete --keep-order --exclusive --command prefligit --arguments "(COMPLETE=fish "'prefligit'" -- (commandline --current-process --tokenize --cut-at-cursor) (commandline --current-token))"
```

> It doesn’t let package managers like Homebrew create those completions while building and pass them on to the user, since the binary’s path might differ depending on the user’s setup.

At least for Homebrew, this is fine since we build binaries per supported prefix (`/opt/homebrew` on mac and `/home/linuxbrew/.linuxbrew` on linux) so the build machine's path will be the same. [jj](https://github.com/Homebrew/homebrew-core/blob/45256b12e948ed1bb7dd0a43007aee9b23d46483/Formula/j/jj.rb#L24) for example uses clap's dynamic completions.

> Do you think it’s good for us to ship it as is and find a better way to integrate it with generate-shell-completion later?

Yea I think we can go ahead and ship this as is

PR #383, opened 2025-08-07
Implement `pygrep` language support

### LINKED ISSUE #38: Support `pygrep` language

j178:
Closes #38


j178:
This is tougher than I thought—maybe even impossible to do in pure Rust. There’s just too many differences in how regex works between Python and Rust.

PR #392, opened 2025-08-07
Refactor `fix_end_of_file` to use BufReader and BufWriter

xomvio:
This PR refactors the fix_file function in fix_end_of_file.rs. Used BufReader and BufWriter for more efficient IO especially dealing with large files. Implementation avoids to read whole file into memory. I will also edit the fix_trailing_whitespace with similar approach. Related to #294.

I know the code is confusing but it's does all the operation at once, instead of re-reading the file. I tried to give comment lines as much as possible to make it clear.

## Changes
* Rewrote `fix_file` using `tokio::BufReader` and `BufWriter` for efficient byte-level processing.
* Introduced `ReverseReader` struct to make it start reading from last byte and continue backwards.
* Rewrote newline trimming logic to work in a streaming manner without full-buffer allocation.
* Removed `detect_line_ending`. The new implementation stops at the first detected newline and uses that style consistently. This approach is commonly used by other projects (though they read forward instead of backward).
* Changed the default line ending to CRLF instead of LF when no newline is present in the file. This can be easily changed to LF if needed. I chose CRLF for better cross-platform compatibility, especially on Windows.

## Note
The new code passes `test_mixed_line_endings`, but this is coincidental. It defaults to using `\r\n` as the line ending, which happens to match what's already common in the test content. However, if the input text were different, this test could fail. Not due to a bug, but due to a change in behavior.

I believe this test is no longer meaningful under the new logic and should be removed or rewritten accordingly.

### DIFF
--- a/src/builtin/pre_commit_hooks/fix_end_of_file.rs
+++ b/src/builtin/pre_commit_hooks/fix_end_of_file.rs
@@ -1,5 +1,6 @@
 use anyhow::Result;
 use futures::StreamExt;
+use tokio::io::{AsyncReadExt, AsyncSeekExt, AsyncWriteExt, BufReader, BufWriter};
 
 use crate::hook::Hook;
 use crate::run::CONCURRENCY;
@@ -7,9 +8,7 @@ use crate::run::CONCURRENCY;
 pub(crate) async fn fix_end_of_file(_hook: &Hook, filenames: &[&String]) -> Result<(i32, Vec<u8>)> {
     let mut tasks = futures::stream::iter(filenames)
         .map(async |filename| {
-            // TODO: avoid reading the whole file into memory
-            let content = fs_err::tokio::read(filename).await?;
-            fix_file(filename, &content).await
+            fix_file(filename).await
         })
         .buffered(*CONCURRENCY);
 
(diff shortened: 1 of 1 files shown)


xomvio:
is current fix_trailing_whitespace working properly? Can you please tell me what output it should give exactly?

j178:
> is current fix_trailing_whitespace working properly?

Sorry, `fix_trailing_whitespace` is totally untested(I forgeted why...), it might have some issues.

> Can you please tell me what output it should give exactly?

You can try setting `PREFLIGIT_NO_FAST_PATH` to disable the Rust implementation and fall back to the original pre-commit-hooks implementation. The output should match the original implementation.

xomvio:
Hi again, I think #399 does pretty well!

j178:
@xomvio Do you prefer #399? It looks like this one and #399 are doing pretty much the same thing, just with slight differences in how they approach it. I need sometime to take a close look, and maybe merge them to get the best of both. Big thanks to both @xomvio and @tmdhhl for all your hard work -- it's really appreciated!

xomvio:
I think #399 is better than this one. For example, when I change the test_preserve_windows_line_endings like this for speed testing:
```rust
#[tokio::test]
async fn test_preserve_windows_line_endings() {
    let time = std::time::SystemTime::now();
    for _ in 0..1000 {
        let dir = tempdir().unwrap();

        let content = b"line1\r\nline2\r\nline3";
        let file_path = create_test_file(&dir, "windows_no_eof.txt", content).await;

        let (code, output) = run_fix_on_file(&file_path).await;

        assert_eq!(code, 1, "Should fix the file");
        assert!(output.as_bytes().contains_str("Fixing"));

        let new_content = fs_err::tokio::read(&file_path).await.unwrap();
        assert_eq!(new_content, b"line1\r\nline2\r\nline3\r\n");
    }
    println!("Elapsed: {}", time.elapsed().unwrap().as_millis());
}
```
#399 output is nearly 300ms while #392 is 550ms
Also, #399 looks like safer for edge cases.

j178:
Since #399 got merged, I'm going to close this one, hope you find some fun with other issues too :)

PR #399, opened 2025-08-08
Avoid reading whole file into memory in `fix_end_of_file` and make it consistent with `pre-commit-hooks`

tmdhhl:
## Background

The current implementation in this repository for fixing trailing newlines in files (see PR #392, Related to #294) has achieved its functional goals and has a relatively simple overall structure. However, in practical use, I noticed some room for improvement, mainly including:

- **Multiple file openings**: The current approach opens the file in read-only mode first, then reopens it in write mode, which may cause additional system call overhead and potential race conditions on some platforms.
- **Lack of comprehensive newline statistics**: When the file ends without a newline, the method scans backward from the file's end and adopts the first newline it encounters as the standard, without counting newlines throughout the file. This can lead to inconsistencies, especially if multiple newline types appear in the file, since the chosen newline might not be the most frequently used one.
- **No flexible configuration for scanning logic**: The current reading logic does not support customizing the scan block size, limiting adaptability for different scenarios such as very long lines or large files.
- **Repeated newline detection code**: The logic for reading the file and detecting trailing newlines appears in multiple places, increasing maintenance cost.

---

## Improvements in This Submission

To address the above issues, I made the following improvements in this submission:

- ✅ **Open the file only once**: Unified asynchronous file opening, with reading and writing handled through the same file handle, reducing system resource usage and minimizing platform-dependent risks.
- ✅ **Block-wise backward scanning from the tail with newline statistics**: Introduced a "tail scanning" strategy that reads blocks from the end of the file in reverse order, counting different newline types to ensure the most frequently used newline is chosen as the standard, avoiding misjudgment.
- ✅ **Support customizable scan block size**: Uses a default 4KiB buffer, while allowing flexible adjustment via the `max_scan_size` parameter to improve adaptability and extensibility.
- ✅ **Enhance code robustness and cross-platform compatibility**: Avoid unnecessary repeated file openings and improve stability across different operating systems and filesystems.

---

## Comparison

| Aspect             | This Submission (Tail Block Scan)           | PR #392 Implementation                   |
|--------------------|---------------------------------------------|-----------------------------------------|
| File Open Count    | Open once with read/write reuse              | Opened twice, duplicated opening risk   |
| File Reading Method| Block-wise backward scanning with newline counting | Backward scan stops at first newline without full distribution count |
| Scan Granularity   | Supports flexible configuration via `scan_size` | Fixed scan size, lacks flexibility      |
| Code Structure and Complexity | Slightly more complex, well encapsulated, easy to test | Simpler logic, but contains repetition and potential inaccuracies |
| Maintainability    | Clear modular logic, easier to extend        | Repeated logic, higher maintenance cost |
| Platform Compatibility | Broader adaptation, avoids file handle conflicts | Multiple file opens may cause platform-dependent behavior |

---

## Summary

This PR addresses current shortcomings in performance, resource utilization, flexibility, and maintainability by providing a more efficient, robust, and flexible alternative。

### DIFF
--- a/src/builtin/pre_commit_hooks/fix_end_of_file.rs
+++ b/src/builtin/pre_commit_hooks/fix_end_of_file.rs
@@ -65,18 +65,18 @@ impl LineEndingDetector {
         }
 
         let mut pre_tail_buf = [0u8; 1];
-        let mut readed_len = 0;
+        let mut read_len = 0;
         let mut buf = vec![0u8; max_scan_size];
 
-        while readed_len < data_len {
-            let block_size = max_scan_size.min(usize::try_from(data_len - readed_len)?);
+        while read_len < data_len {
+            let block_size = max_scan_size.min(usize::try_from(data_len - read_len)?);
             line_ending_detector
                 .read_bytes_backward(reader, &mut buf[..block_size], false)
                 .await?;
-            readed_len += block_size as u64;
+            read_len += block_size as u64;
 
             // Cache last byte of previous block to avoid splitting `b"\r\n"`.
-            if readed_len != data_len {
+            if read_len != data_len {
                 line_ending_detector
                     .read_bytes_backward(reader, &mut pre_tail_buf, true)
                     .await?;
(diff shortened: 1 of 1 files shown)


j178:
Hi @tmdhhl and @xomvio, I picked up some things from both your PR and simplified the implementation. Plus, I fixed a difference with the original pre-commit-hooks: if there’s no line ending at the end of a file, we just add a "\n" at the end without worrying about the previous line endings. It might seem a bit odd, but I wanna keep things consistent with pre-commit-hooks. 

I also added @xomvio to the author list of this commit, thank you for the constribution!

j178:
Updates #294

PR #402, opened 2025-08-08
Rename to `prek` from `prefligit`

j178:
This is a breaking change:
- The cache dir has changed from `~/.cache/prefligit` to `~/.cache/prek`, all hooks are going to be reinstalled.
- All env var prefix changed to`PREK_` now.
- Homebrew-core formula needs to be updated(or create a new one?)


wey-gu:
- [ ] logo/banner image's text

PR #420, opened 2025-08-11
Use `brew install j178/tap/prek` for now

fan-tastic-z:


j178:
When our name was `prefligit`, it was in homebrew-core. We changed it to `prek`, but the rename PR in homebrew-core is stuck waiting for Rust 1.89 updates. For now, I think we go with `brew install j178/tap/prek`, and once   https://github.com/Homebrew/homebrew-core/pull/232740 gets merged, we’ll switch back to `brew install prek`.

j178:
Thanks!

fan-tastic-z:
> When our name was `prefligit`, it was in homebrew-core. We changed it to `prek`, but the rename PR in homebrew-core is stuck waiting for Rust 1.89 updates. For now, I think we go with `brew install j178/tap/prek`, and once [Homebrew/homebrew-core#232740](https://github.com/Homebrew/homebrew-core/pull/232740) gets merged, we’ll switch back to `brew install prek`.

Got it, thank you

PR #424, opened 2025-08-11
Implement `prek list` to list available hooks

j178:
- **.**
- **Implement `prek list` to list available hooks**


potiuk:
Nice !

PR #437, opened 2025-08-12
Add check for `minimum_prek_version`

### LINKED ISSUE #421: Implement minimum-prek-version
Similarly to pre-commit, It would be great to have `minimum-prek-version` in the config file - which would fail prek if the version is not high enough. 

```
default_language_version:
  python: python3
  node: 22.18.0
  golang: 1.24.0
minimum_prek_version: '0.0.23'
exclude: ^.*/.*_vendor/
```

I think that is quite **crucial** to be implemented before first version considered "ready" - because there is a small problem if it is not implemented and checked - imagine that 0.30.0 is released and we put `0.30.0` in there, but if someone has `0.23.0` they might be stuck with that version for quite some time because 0.23.0 does not have that check :)

You need to have a version that checks it installed for it to be effective - so the sooner such check is implemented, the more effective it is ;)

j178:
Closes #421


potiuk:
Cool!

PR #444, opened 2025-08-13
Implement `mixed_line_ending` as built hook

### LINKED ISSUE #412: Implement pre-commit-hooks:mixed-line-ending
The original implementation: https://github.com/pre-commit/pre-commit-hooks/blob/main/pre_commit_hooks/mixed_line_ending.py

lemorage:
Resolve #412

### DIFF
--- a/src/builtin/pre_commit_hooks/mixed_line_ending.rs
+++ b/src/builtin/pre_commit_hooks/mixed_line_ending.rs
@@ -25,7 +25,7 @@ pub(crate) async fn mixed_line_ending(
     hook: &Hook,
     filenames: &[&String],
 ) -> Result<(i32, Vec<u8>)> {
-    let fix_mode = parse_fix_mode(&hook.args);
+    let fix_mode = parse_fix_mode(&hook.args)?;
 
     let mut results = futures::stream::iter(filenames)
         .map(|filename| fix_file(filename, &fix_mode))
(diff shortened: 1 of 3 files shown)


src/builtin/pre_commit_hooks/mixed_line_ending.rs
```diff
@@ -0,0 +1,295 @@
...
+
+    let (exit_code, output) =
+        results
+            .into_iter()
+            .try_fold((0, Vec::new()), |(mut code, mut output), result| {
+                let (file_code, file_output) = result?;
+                code |= file_code;
+                output.extend(file_output);
+                Ok::<_, anyhow::Error>((code, output))
+            })?;
```
j178: nit: can we use a plain `for-loop` instead?

src/builtin/pre_commit_hooks/mixed_line_ending.rs
```diff
@@ -0,0 +1,295 @@
...
+
+    Ok((exit_code, output))
+}
+
+// Parse the fix mode from command line arguments
+fn parse_fix_mode(args: &[String]) -> FixMode {
```
j178: Looks like it doesn’t handle arguments like `["--fix", "no"]`, right?

j178 (commented):
Thank you! Added two little comments.

src/builtin/pre_commit_hooks/mixed_line_ending.rs
lemorage: Added!

tests/common/mod.rs
```diff
@@ -233,6 +233,17 @@ impl TestContext {
             .success();
     }
 
+    /// Configure git autocrlf.
+    pub fn configure_git_autocrlf(&self) {
```
j178: Can we name it `disable_git_autocrlf`?

lemorage:
Thanks for the changes! I was looking at it.

PR #463, opened 2025-08-15
warn unexpected keys.

### LINKED ISSUE #435: Add unexpected additional keys warning
https://github.com/pre-commit/pre-commit/blob/b74a22d96cca546b8e0bb9f68f1d7d8565205b65/pre_commit/clientlib.py#L469C5-L482C7

tmdhhl:
closed #435


tests/validate.rs
```diff
@@ -116,3 +116,32 @@ fn validate_manifest() -> anyhow::Result<()> {
...
+                name: Test Hook
+                entry: echo test
+                language: system
+                unexpected_key_in_hook: some_value
+        unexpected_key: some_value
+        another_unknown: test
+        minimum_pre_commit_version: 1.0.0
+    "});
+
+    // TODO: warning about `unexpected_key_in_hook` currently not working
```
j178, line 138: @tmdhhl Hi, any thoughts on how to make this work? If not, we can just merge it now.

tmdhhl: Actually, serde_ignored already detects extraneous keys at different levels — we just need to add the handling logic for hooks.[ demo](https://docs.rs/serde_ignored/latest/serde_ignored/)

tmdhhl: > @tmdhhl Hi, any thoughts on how to make this work? If not, we can just merge it now.

`serde_ignored`cannot correctly detect extraneous keys when flatten is used. We need some time to think about how to properly handle the detection of fields under hook.

j178: Thanks for confirmation, created https://github.com/j178/prek/issues/470 to track, this PR is good enough to merge I think.

PR #464, opened 2025-08-15
Download uv from github releases directly

j178:


src/languages/python/uv.rs
```diff
@@ -64,6 +65,25 @@ fn get_platform_tag() -> Result<String> {
     Ok(platform_tag.to_string())
 }
 
+/// Get the GitHub platform string and archive extension for the current host.
+fn get_github_platform_info() -> Result<(&'static str, &'static str)> {
+    let os = std::env::consts::OS;
+    let arch = std::env::consts::ARCH;
+
+    // Returns (platform_string, archive_extension)
+    let (platform, extension) = match (os, arch) {
```
j178: add more platforms

PR #469, opened 2025-08-15
Propagates color choice env variables to hooks and commands

potiuk:
This change propagats color choice of prek to the hooks and commands it executes. It does it by setting the environment variables based on Always/Auto/Never.

For now we are using just pretty "standard" env variables used in a number of tools:

* NO_COLOR -> obvious (Never)
* FORCE_COLOR -> when you want to force the color regardless of detection (used for example by ruff)a (Always)
* CLICOLOR -> when you want to "hint" to other rust binaries that the color is available (used by colored crate) (Auto)
* CLICOLOR_FORCE -> when you want to force setting color for coloured crate supported binaries (Auto)

While this does not solve ruff colouring when pseudo terminal is not allocated, it does color the output even if there is no pseudo-terminal.

With this change, for example ruff output will be coloured when --color always is used (which did not work before) - you had to manually set FORCE_COLOR=1 before running prek to make it works..

This is not a complete solution yet. The problem currently is that we are capturing output from the commands that we execute. Even if we allocate pseudo terminal and pass it as "usual" this will not solve the problem because pseudo terminal in this case will be only used by stdin and capturing the output will cause the stdout and stderr to be pipes (i.e, isatty returns False). What we really want in the final solution is to create pseudo terminal, pass it (including stdout and stderr) to the command, and then read the output and error from the pseudo terminal. This will make it possible for auto detection of colour by all tools to work properly.


potiuk:
Hey @j178 - I've added `pty-check` prek hook - this makes it much easier to test locally what terminal information hooks receive.

Sorry if that's a little rough at the edges, but this is my first rust code attempt :)

potiuk:
I'v added tests as well :)

potiuk:
I also changed the `pty-check` pre-commit to be `manual` stage and `verbose` by default. This makes it a little easier to use it for testing, but also it makes it easily possible to switch and test how terminal does not get allocated when `git commit` is run - you will see all the terminal information when you run `git commit` if you add the stage.

potiuk:
Also added more description on the purpose of the `pty-check` prek hook in the comment.

I think it's generally ready for review :)

potiuk:
And BTW. Even if just **this** one is merged, it should now be possible to switch airflow to prek. I will just add `--color always` to our prek run in CI, and add some description of using `--color always` locally to see colors or setting `FORCE_COLOR=1` env variable (both works). But I am also looking at implementing the pseudo-terminal as well. I already have working code on allocating `pty` (based on `portable-pty` crate) , but what I need to do is to make this little dance of not capturing the output in the command, but rather read it from the pseudo-terminal - currently we are running .output() on the command, which automatically sets stderr and stdin to be pipes, and that makes them non-TTY stdout/stderr  - so auto-detection of color does not work for ruff for example.

j178:
Great to see this! `pty_check.py` is definitely helpful for testing!

Just a heads up, I’m also playing around with `pty-process` crate for a real PTY solution. I think it’s better since it has async support and works with tokio, which makes it easier to use in our async code.

Passing color choices to the child process might not be as important once we’re there? but I’ll think about it. If we pull out the `pty_check.py` script into a separate PR, we can probably get it merged pretty quickly.

potiuk:
I think it's worth to merge it regardless (that's why I proposed it) - because I think it's nice that `--color` flag gets nicely propagated to the hooks anyway. this way with a single flag on `prek` you effectively disable or force colors for **everything** - PTY is good for auto-detection for sure, but ability to control the color for "prek + hooks" with a flag is I think pretty useful.

potiuk:
But - it's up to you of course - I can just leave the pty-check :) or tweak the vars - for example the 


```
            ColorChoice::Auto => {
                self.env("CLICOLOR", "1");
            }
```

Is a bit that I was not sure about.

potiuk:
> Just a heads up, I’m also playing around with pty-process crate for a real PTY solution. I think it’s better since it has async support and works with tokio, which makes it easier to use in our async code.

That's cool. I can play as well - I am not afraid of duplicating this work, I treat it also as a learning exercise for me and in the worst case it will make my reviews of your code way better and more informed as I will get the hang of rust a little more (I already am)

potiuk:
Yeah. I thought a  bit and indeed - closing it is better. I think there is already an easy way to for color/no color - by just setting the env variables before runnig prek - and they will affect both prek and the hooks, so there is no point really to apply some special behaviour to `--color` -> it can **only** refer to `prek` and not to the hooks.

Just before closing this one I looked at the terminal information that hooks get with check-pty  and it looks as expected. So the `check-pty` is no longer needed. 

BTW. It was purely vibe-coded of course - I just asked copilot to generate python script showing terminal info and env variables :). 

Regular prek run:

Screenshot 2025-08-16 at 11 03 44

Prek run inside git:

Screenshot 2025-08-16 at 11 04 56

PR #481, opened 2025-08-17
Support multiple `.pre-commit-connfig.yaml` in a workspace (monorepo mode)

j178:
- **Refactor project constructure**
- **.**
- **.**
- **Support multiple `.pre-commit-connfig.yaml` in a workspace (monorepo mode)**


j178:
@potiuk This basicly works now, you may want to play with it :)

```console
❯ tree -a -I .git
.
├── .pre-commit-config.yaml
├── a.py
├── backend
│   └── .pre-commit-config.yaml
└── frontend
    ├── .pre-commit-config.yaml
    └── hh.txt

❯ ~/code/rust/prek/target/debug/prek
Running hooks from `.pre-commit-config.yaml`:
script...................................................................Passed

Running hooks from `backend/.pre-commit-config.yaml`:
trim trailing whitespace.................................................Passed
fix end of files.........................................................Passed
check yaml...............................................................Passed
check for added large files..............................................Passed

Running hooks from `frontend/.pre-commit-config.yaml`:
check for added large files..............................................Passed
fix end of files.........................................................Passed
check yaml...............................................................Passed
trim trailing whitespace.................................................Passed
```

potiuk:
Wowowow :)

kaxil:
Nice, I was missing this in pre-commit since Airflow is a mono-repo (and a few more mono-repos that I work with at Astronomer)!

potiuk:
Quite a few comments (WIP/Draft of separating golang) : https://github.com/apache/airflow/pull/54615

Bugs: 

* the monorepo support does not detect our `.github/.pre-commit-config.yaml` (we have separate octopin hooks there - as it needs python 3.11
* the pre-commits from golang one are detected for `prek run` but they are not detected for regular "git commit"
* seems that docker pre-commit run from sub-dir stil gets main directory mounted as /src/ and not the subdir - this is why I have `entry: -w /src/go-sdk vektra/mockery:3` - this should ideally be `-w /src vektra/mockery:3. But we were at least able to switch to using golang type for `go mod tidy` which is quite cool (no need to have go installed locally for it), gci nicely detects the prefixes so I could simplify it :) ... That's cool.
* auto-complete does not show those prek hooks coming from the sub-dirs
* prek list does not show the pre-commits from sub-dirs

Some usability improvement proposals:

* it would be cool if prek run with working dir iin  subdir only runs it on the subdir .pre-commit-config.yaml. This way `cd go-sdk` and you only work on golang things - nothing else. Currently it runs hooks from all detected files. This also should play nicely with the docker mapping of /src above. Generally (except the .git information and folder retrieval) `prek` that is run while inside the `go-sdk` folder should run as if `go-sdk` was the ROOT of the repository.  This will have some consequences for `--config` - but still I think it makes sense that if your wd contains .pre-commit-config.yaml - that should be the dir that prek considers as ROOT (and maps to docker hooks as /src for example).
 
Cool would be features:

* would be great to be able to specify just `go-sdk` hooks to run from the main folder (but it somehow clashes with --directory flag, you'd have to likely play a bit with names of flags): `--hook-directory` ??? or smth.
* would be great to namespace the hooks with the directory prefix. I think it would make much more sense if the hooks from go-sdk are automatically named `go-sdk/hook-name`. That would play very nicely with auto-commit  and it would allow to avoid any kind of name clashes. Also that would allow `prek run go-sdk/*` for example as equivalent of `--hook-directory` or even `prek run go-sdk/go*` as an interesting way of running only subset of `prek` hooks

I did not split it in several issues as I think this is more of a general feedback and you can pick and choose how to approach it.

potiuk:
Another comment. 

Thinking about our CI usage - we would also probably like to be able to disable the workspace feature and only run `prek` using the current `.pre-commit-config.yaml` - and not detect any other config files.

This would enable us to have much more modular CI run - and have seaprate jobs for each .pre-commit-config.yaml . 

There might be multiple reasons for that:

* like for example today we do not run `.github/.pre-commit-config.yaml` in the same `prek` run as the main one - because it requires python 3.11 and we want to run all other pre-commits in main with 3.10 - to make sure that it works with the "lowest" version of Python we support.

* another reason for that  - it's just nicer to have separate "golang" static checks and "python" static checks and later "helm static checks" and "whatever else static checks" as separate jobs - this is much easier to look through much smaller list of hooks run and it immediatly shows which part failed rather than "Generic static check failed" which we currently have and it is not telling too much until you open the log and check (and often that log gets long when there are a lot of failures and github UI is not very friendly with long logs.

potiuk:
From our CI now:

Screenshot 2025-08-18 at 16 18 02

potiuk:
BTW: Would be nice to have just (in the output):

* Running hooks from`go-sdk`
* Running hooks from`.`

potiuk:
Small update from @ashb -> https://github.com/apache/airflow/pull/54615#discussion_r2282921798 looks like the golang "working dir" is also still not the right one, i.e. golang hooks run for the sub-dir hook have the main repo folder as working dir. Which is consistent with what I saw in `docker`

j178:
I’ve already got the basic support in place, but this PR is huge and has a ton of random changes. So, I’m gonna merge it into the `workspace` branch and then keep working on more specific features there.

potiuk:
All my issues are now fixed :)

Issue #434, opened 2025-08-12
Add moving `rev` warning


j178:
https://github.com/pre-commit/pre-commit/blob/b74a22d96cca546b8e0bb9f68f1d7d8565205b65/pre_commit/clientlib.py#L247-L263

luizvbo:
WIP: https://github.com/j178/prek/pull/488

PR #488, opened 2025-08-18
Add moving rev warning

luizvbo:


luizvbo:
Closes https://github.com/j178/prek/issues/434

j178:
Thanks! But I think what we what is a simple regex check?

This is the original pre-commit implementation:
```py
class WarnMutableRev(cfgv.Conditional):
    def check(self, dct: dict[str, Any]) -> None:
        super().check(dct)

        if self.key in dct:
            rev = dct[self.key]

            if '.' not in rev and not re.match(r'^[a-fA-F0-9]+$', rev):
                logger.warning(
                    f'The {self.key!r} field of repo {dct["repo"]!r} '
                    f'appears to be a mutable reference '
                    f'(moving tag / branch).  Mutable references are never '
                    f'updated after first install and are not supported.  '
                    f'See https://pre-commit.com/#using-the-latest-version-for-a-repository '  # noqa: E501
                    f'for more details.  '
                    f'Hint: `pre-commit autoupdate` often fixes this.',
                )
```

luizvbo:
Ok. I'll see if I can find some time tomorrow to implement the other approach

luizvbo:
@j178 I changed according your suggestion. Wdyt?

j178:
Thanks! I adjusted the warning message a bit to cut down on the noise when a configuration has several moving repos.

PR #489, opened 2025-08-18
Build wheels for more platforms

### LINKED ISSUE #451: Publish more platforms in `pypi`
Prek currently publishes small number of binary wheels in `pypi` and it means that on a number of platforms you'd have to have `build-essentials` to compile prek from .sdist - not mentioning potential dependency on installed libraries needed for that (such as .xz). I think there should be more platforms available to increase the easyness of installing prek from `pypi` for wider audience.

Comparing `prek`:

Image

To `uv` (which could be considered as a baseline):

Image

j178:
Closes #451


potiuk:
Cooool !

kaxil:
That was fast :) thanks @j178

PR #523, opened 2025-08-20
add OpenLineage to prek users

mobuchowski:
We use prek at OpenLineage, thanks for @JDarDagran for setting this up. https://github.com/OpenLineage/OpenLineage/pull/3965


j178:
Thanks, great to hear that!

potiuk:
Spreading like fire !

PR #540, opened 2025-08-25
Sensible regex warning

luizvbo:


j178:
Hi, thanks for your contribution as always! Could you let me know why this was closed?

luizvbo:
I thought this issue would be easy to fix, but since the checking has to be done before the SerdeRegex is applied, it's a bit more complex.

PR #549, opened 2025-08-26
improve: avoid unnecessary alloc

dyxushuai:


j178:
Oh my god, I didn’t expect the fix could be that simple, thanks!

src/run.rs
```diff
@@ -151,12 +151,9 @@ where
...
     );
 
+    #[allow(clippy::redundant_closure)]
     let mut tasks = futures::stream::iter(partitions)
-        .map(|batch| {
-            // TODO: avoid this allocation
-            let batch: Vec<_> = batch.iter().map(ToString::to_string).collect();
-            run(batch)
-        })
+        .map(|batch| run(batch))
```
j178, line 156: That’s a bit beyond my Rust knowledge. When I remove the redundant closure, it cannot compile 😞

dyxushuai: The bound is `F: for<'a> AsyncFn(&'a [&'a String]) -> anyhow::Result<T>`, but `map` require: `F: FnMut(Self::Item) -> T`, so we need to indirect call `|batch| run(batch)` to satisfy the bound

dyxushuai: If you want to change the bound of `F`, the announcement may become a bit more complex, like this:
`F:  for<'a> FnMut(&'a [&'a String]) -> Fut, Fut: Future<Output=anyhow::Result<T>>`

j178: Thanks for explaining it—I learned something new today!

PR #557, opened 2025-08-26
Implement `check_yaml` as builtin hook

### LINKED ISSUE #531: feat: support check-yaml as a builtin hook

lemorage:
Close #531

### DIFF
--- a/src/builtin/mod.rs
+++ b/src/builtin/mod.rs
@@ -19,7 +19,10 @@ pub fn check_fast_path(hook: &Hook) -> bool {
             if *NO_FAST_PATH {
                 return false;
             }
-            Implemented::from_str(hook.id.as_str()).is_ok()
+            let Ok(implemented) = Implemented::from_str(hook.id.as_str()) else {
+                return false;
+            };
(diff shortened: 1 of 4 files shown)


j178 (commented):
Thanks! I’m a bit worried because we’re missing the `--unsafe` and `--allow-multiple-documents` features, and I know it’s pretty tough to implement these exactly in Rust.

I’m thinking we should only use our built-in implementation if the c`heck-yaml` hook doesn’t include `--unsafe` or `--allow-multiple-documents` arguments. Basically, we should improve our detection in the `check_fast_path` logic for better compatibility. What do you think?

lemorage:
Fair considerations to me. But simply adding a special case to `check_fast_path` makes things bloated, also is not quite clean and extensible. If this is not a burden on your side, I can make this addition later.

j178:
I added a basic check over at https://github.com/j178/prek/pull/557/commits/dc1ccf79f9a45550ee9b624257e59476fdb235ec, I think it's fine for now.

lemorage:
Thx!

PR #581, opened 2025-09-01
Add `prek run --cd <dir>` to change directory before running

j178:
- **Support multiple `.pre-commit-config.yaml` in a workspace (monorepo mode) (#481)**
- **Add `prek run --cd <dir>` to change directory before running**


potiuk:
Nice. Good idea to split what's needed into separate PRs :)

PR #582, opened 2025-09-01
Add an full example and update docs

j178:


potiuk:
Nice! @iamhatesz - as discussed at Pycon PL - the monorepo support is shaping up nicely with all the work @j178 puts in it :)

iamhatesz:
Thanks @potiuk. The ticket to migrate is already on our board. :)

PR #677, opened 2025-09-08
Ignore config file from hidden directory

j178:


potiuk:
That is a bit problematic for us as we have a .pre-commi-config.yaml already in .gitHub - but I guess it's a good call, we can move it

j178:
> we can move it 

@potiuk This PR stops auto-discovery of the .pre-commit-config.yaml file in hidden directories. Having it there makes the directory a project, but that doesn’t seem right for something like .github, does it? I think it’s always used with the -c option, so this PR won’t affect that.

potiuk:
> > we can move it
> 
> @potiuk This PR stops auto-discovery of the .pre-commit-config.yaml file in hidden directories. Having it there makes the directory a project, but that doesn’t seem right for something like .github, does it? I think it’s always used with the -c option, so this PR won’t affect that.

No worries. I already moved it. We **do** want this check to be run at `git commit` but it can be easily done in other non-hidden repository. The only reason we separated it originally (without monorepo support) was that it required Python 3.11 due to octopin needing it, and all the rest of our hooks are Python 3.10+ (because all our code is Python 3.10+) - but if we could run it from main .pre-commit-config.yaml, we would. 

Now with monorepo support we **want** to run it automatically on `git commit` - but it's perfectly fine to move it to another directory as well. Not following hidden directories makes it much more performant to search to through checked out repo, so it's a good call I think.

PR #687, opened 2025-09-09
docs(manifest): Correctly specify metadata for all packages

epage:
- Having `repository` on every package will be helpful for provenance checking, see also https://lawngno.me/blog/2024/06/10/divine-provenance.html
- All packages should have a license
- `package.license` is the correct field to use for standard licenses


j178:
Thank you!

PR #692, opened 2025-09-09
Fix Docker image and Go version test issues for portability

aravindan888:
Fixes#690 

This PR resolves two test failures that occurred on machines using Podman or with multiple Go versions:
docker_image.rs test – Updated to use fully qualified Docker image names (docker.io/zricethezav/gitleaks:v8.21.2) so the test works reliably on both Docker and Podman. Previously, short image names weren’t automatically resolved, causing code=125 errors on Podman.
golang.rs test – Modified the Go version assertion to check that 1.23.11 exists rather than expecting it to be the only installed version. This prevents test failures on systems that have multiple Go versions installed

### DIFF
--- a/tests/languages/golang.rs
+++ b/tests/languages/golang.rs
@@ -2,8 +2,6 @@ use assert_fs::assert::PathAssert;
 use assert_fs::fixture::PathChild;
 
 use crate::common::{TestContext, cmd_snapshot};
-
-// We use `setup-go` action to install go1.24.5 in CI, so 1.23.11 should be downloaded by prek.
 #[test]
 fn language_version() -> anyhow::Result<()> {
     let context = TestContext::new();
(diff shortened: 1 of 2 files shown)


tests/languages/golang.rs
```diff
@@ -102,17 +100,21 @@ fn language_version() -> anyhow::Result<()> {
...
-            .home_dir()
-            .join("tools")
-            .join("go")
-            .read_dir()?
-            .flatten()
-            .filter(|d| !d.file_name().to_string_lossy().starts_with('.'))
-            .map(|d| d.file_name().to_string_lossy().to_string())
-            .collect::<Vec<_>>(),
-        vec!["1.23.11"],
+    // ✅ Updated assertion
```
j178:
```suggestion
```

j178:
Thank you!

aravindan888:
this should work, good to merge

PR #713, opened 2025-09-11
Speed up check_toml

### LINKED ISSUE #710: Potential speed up for `check_toml`
It appears the built-in `check_toml` parses the TOML into an owned type.
https://github.com/j178/prek/blob/3d6f9f1d5a04cc26d09efc58815e85931b1a77bd/src/builtin/pre_commit_hooks/check_toml.rs#L32-L38

with toml v0.9, you can now use [`toml::de::DeTable::parse`](https://docs.rs/toml/latest/toml/de/type.DeTable.html#method.parse) to parse into borrowed types, avoiding extraneous memory allocations.

Enabling the `preserve_order` feature can further cut parse times in half because of the performance difference between `BTreeMap` (the default) and `IndexMap` (toml-rs/toml#945).  `fast_hash` builds on that and makes things even faster (toml-rs/toml#956).

kushudai:
closes #710

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,7 +12,7 @@ tracing = { version = "0.1.40" }
 
 [package]
 name = "prek"
-version = "0.2.0-alpha.4"
+version = "0.2.0-alpha.5"
 authors = ["j178 <hi@j178.dev>"]
 description = "Better `pre-commit`, re-engineered in Rust"
 repository = { workspace = true }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,28 @@
 # Changelog
 
+## 0.2.0a5
+
+### Enhancements
+
+- Add built in byte-order-marker fixer ([#700](https://github.com/j178/prek/pull/700))
+- Use bigger buffer for fixing trailing whitespace ([#705](https://github.com/j178/prek/pull/705))
+
+### Bug fixes
+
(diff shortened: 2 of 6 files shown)


j178:
Thanks, this is fast!

kushudai:
> Thanks, this is fast!

Haha, thanks I have worked toml a lot recently :)

Cargo.toml
```diff
@@ -83,7 +83,7 @@ textwrap = { version = "0.16.1" }
 thiserror = { version = "2.0.11" }
 tokio = { version = "1.47.1", features = ["fs", "process", "rt", "sync", "macros"] }
 tokio-util = { version = "0.7.13" }
-toml = { version = "0.9.5", default-features = false, features = ["parse", "serde"] }
+toml = { version = "0.9.5", default-features = false, features = ["fast_hash", "parse", "preserve_order", "serde"] }
```
j178, line 86: Do you know if `fast_hash` and `preserve_order` are required? I noticed they’re bringing in some new dependencies.

kushudai: Both are speed optimizations as @epage mentioned in #710 
If you'd like to avoid the deps, I can remove them. `preserve_order` brings in `indexmap` and `fast_hash` brings that + `foldhash`

kushudai: Let me know if I should remove @j178 - otherwise this is good to review :)

j178: `indexmap` is already in the deps tree, so only `fast_hash` is new, I think it is good to keep.

PR #727, opened 2025-09-14
Fix typo in language-version to language_version

onerandomusername:
Checking the pre-commit documentation and the source code of prek will reveal that this is named language_version rather than language-version. 

It's a tiny nit but a little confusing otherwise


j178:
Thanks!

PR #731, opened 2025-09-14
Use concurrent read and write in `git check-attr`

### LINKED ISSUE #730: `check_added_large_files` hangs when used in a repository with too many files
When I run `prek run  -a` in an existing repository prek hangs. Partial output from `ps` follows:

```
matthias   48903  0.0  0.0 235744  8524 pts/5    Ss   10:33   0:00  |   \_ -bash                                                                                        matthias   54447  1.2  0.0 827340 12440 pts/5    Sl+  10:55   0:00  |   |   \_ prek run -a                                                                              matthias   54472  0.0  0.0 237504  5828 pts/5    S+   10:55   0:00  |   |       \_ /usr/bin/git -c core.useBuiltinFSMonitor=false check-attr filter -z --stdin   
```

`prek` spawns `git check-attr filter` which doesn't seem to do anything. I think this happens when the repository contains too many files but I'm not sure. Unfortunately, I cannot share the repository. I see the same symptoms when running the following script though:

```bash
git init
for dir in $(seq 100) ; do (mkdir $dir; cd $dir ; for f in $(seq 100) ; do (touch "file$f.txt") ; done ) ; done
git add .
prek install
cat > .pre-commit-config.yaml <<EOD
repos:
  - repo: https://github.com/pre-commit/pre-commit-hooks
    rev: v6.0.0
    hooks:
      - id: check-added-large-files
EOD
git commit -m "Stuff"
```

matthiask:
Closes #730 

I'm sure the code is bad, but it passes the test repository in #730 

```
[14.09. 11:41:59] ~/tmp$ time ~/projects/prek/target/debug/prek 
check for added large files..............................................Passed

real    0m0.176s
user    0m0.247s
sys     0m0.140s
[14.09. 11:44:47] ~/tmp$ 
```


matthiask:
The code is possibly less bad now, tests are passing and running the resulting `prek` on my test repository works too. 

That being said, this is the first time I have submitted rust code for real and I did heavily use a coding assistant for this. If the code is crappy that's my fault, but I want to be transparent here.

matthiask:
It seems using `try_join!` might be more idiomatic than `loop { select! { ... }}`

I'll wait for some feedback now :)

src/git.rs
```diff
@@ -422,33 +424,58 @@ pub(crate) async fn get_lfs_files(paths: &[&Path]) -> Result<Vec<String>, Error>
...
+                } else if let Some(path) = path_iter.next() {
+                    if let Some(ref mut stdin_handle) = stdin {
+                        stdin_handle.write_all(path.to_string_lossy().as_bytes()).await?;
+                        stdin_handle.write_all(b"\0").await?;
+                    }
+                    Ok(())
                 } else {
-                    None
+                    stdin.take();
+                    write_complete = true;
```
kushudai: This may be incorrect if git has not written any output after the last one but we've processed all it has written so far?

src/git.rs
```diff
@@ -422,33 +424,58 @@ pub(crate) async fn get_lfs_files(paths: &[&Path]) -> Result<Vec<String>, Error>
...
+                }
+            }
+
+            write_result = async {
+                if write_complete {
+                    std::future::pending::<Result<(), std::io::Error>>().await
+                } else if let Some(path) = path_iter.next() {
+                    if let Some(ref mut stdin_handle) = stdin {
+                        stdin_handle.write_all(path.to_string_lossy().as_bytes()).await?;
+                        stdin_handle.write_all(b"\0").await?;
```
kushudai: We probably want to be batching these writes

kushudai (commented):
I suspect this actually fixes the problem fwiw.
It may pass your repository because the std out was the problem in that case but this can happen if the input we pass to git is also too large since most OSes impose a limit.

We likely need to batch cfg

j178:
@matthiask Thanks for working on this! I put together a version using `tokio::try_join!`, but I’m not familiar with async Rust too. Maybe @kushudai could take a look and review it?

src/git.rs
matthiask: I'm not sure I understand? This is the `stdin` side, and we should be in control there, not git? We know if we still have paths to write, and only set `write_complete` to true afterwards.

(Thinking.) Or maybe you're referring to `0 if write_complete => break` above; that's a good point.

In any case, this problem should be gone in the new `try_join!` version. There, we're using `read_to_end` which hopefully handles this transparently for us.

src/git.rs
matthiask: Good point. I'm not sure we have to do it right now, `prek` is still really fast in my test repo with 10'000 files.

matthiask:
Thank you both!

@j178 I compiled the code and ran it in my test repo, it seems to work fine!

@kushudai It's my understanding that the `loop { select! { ... } }` construct concurrently writes as long as the pipe buffer isn't full and reads as long as there's something to read.

I didn't debug if the problem was on the `stdin` or the `stdout` side of things. I only checked that the problem appears somewhere between a few dozen KiB and about a hundred KiB of input/output. While researching this I stumbled upon 64KiB pipe buffers, which seem to be standard on Linux systems.

matthiask:
🥳

src/git.rs
kushudai: That is indeed what I was referring to :)
Sorry should have left a comment on that line

kushudai:
Apologies for the lateness, I took a look, it looks great. There is more scope for improvement by batching the writes, I may take a stab at that :)

Issue #718, opened 2025-09-11
prek hangs under Zed


j178:
https://github.com/zed-industries/zed/issues/37293

I can consistently reproduce this under Zed, it's super odd: when commiting in Zed, each time a random git process started by prek becomes `<defunct>`, prek keeps waiting for the process to finish, but it never does, so everything just hangs.

CharlesChen0823:
hang with call in [`self.output().await`](https://github.com/j178/prek/blob/2090c5d363f89d53c7a3c376a6eb18142913ce01/src/process.rs#L196), only mac...

j178:
Yes, I tried to replace `.output()` with `.spawn() + tokio::time::sleep() + .wait_with_output()` and it seems doesn't hang anymore

j178:
This isn't just a prek sepcific issue, I can make a minimal reproducer:

```console
$ cargo init zed-reproducer
$ cd zed-reproducer
$ cargo add tokio -F full
```

Edit `src/main.rs`:

```rust
use tokio::process::Command;

#[tokio::main]
async fn main() {
    let output = Command::new("git").arg("version").output().await.unwrap();
    println!("{}", String::from_utf8_lossy(&output.stdout));
}
```

```console
$ cargo build
$ cp ./target/debug/zed-reproducer .git/hooks/pre-commit
```

Open this project in Zed, and commit in Git Panel, then the commit hangs.

CharlesChen0823:
I'm also not sure this is bug in zed or tokio? because in linux I test work fine.

cscnk52:
Windows is ok too, I think this is a bug related to macOS.

CharlesChen0823:
this bug could be fixed by zed in [commit](https://github.com/zed-industries/zed/commit/3780fe3b8ed0c7d4255b54151fcd597010d1f844), @j178 you can testing in zed main branch.

j178:
@CharlesChen0823 Great! I'll wait for the new Zed release to try it out.

By the way, do you happen to know why https://github.com/zed-industries/zed/commit/3780fe3b8ed0c7d4255b54151fcd597010d1f844 is causing this issue? I didn't see much explanation in the commit, and I'm really curious to understand what's going on.

CharlesChen0823:
Sorry, this is just my guess. I really don't known core problem.

j178:
I tried Zed 0.208.6 and 0.209.2-pre, but unfortunately this issue is still happening.

CharlesChen0823:
[...]

PR #756, opened 2025-09-16
fix: rare race condition

bxb100:
close: #718

### DIFF
--- a/src/process.rs
+++ b/src/process.rs
@@ -160,24 +160,46 @@ impl Cmd {
         Ok(())
     }
 
-    /// Equivalent to [`std::process::Command::spawn`][],
+    /// Equivalent to [`tokio::process::Command::spawn`][],
     /// but logged and with the error wrapped.
     pub fn spawn(&mut self) -> Result<tokio::process::Child, Error> {
         self.log_command();
-        self.inner.spawn().map_err(|cause| Error::Exec {
+
+        // temporarily block SIGCHLD on macOS to avoid race conditions
+        // see https://github.com/tokio-rs/tokio/pull/6953
+        #[cfg(target_os = "macos")]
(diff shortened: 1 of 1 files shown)


j178:
Thanks, I can confirm the hang goes away with this fix!

I’m curious though - why does this happen 100% of the time in Zed but never in other situations? Do you have any idea what might be causing that?

bxb100:
This demo in Zed will fail when set as a `pre-commit` hook. The `SIGCHLD` signal is not handled; furthermore, nothing I know.

```rust

use std::sync::{
    Arc,
    atomic::{AtomicUsize, Ordering},
};

extern crate signal_hook_registry;

fn main() {
    let called = Arc::new(AtomicUsize::new(0));

    let hook = {
        let called = Arc::clone(&called);
        move || {
            called.fetch_add(1, Ordering::Relaxed);
        }
    };

    unsafe {
        // let mut mask: libc::sigset_t = std::mem::zeroed();
        // libc::sigemptyset(&raw mut mask);
        // libc::sigaddset(&raw mut mask, libc::SIGCHLD);
        // libc::sigprocmask(libc::SIG_BLOCK, &raw const mask, std::ptr::null_mut());
        let _ = signal_hook_registry::register(libc::SIGCHLD, hook.clone()).unwrap();
        // let mut mask: libc::sigset_t = std::mem::zeroed();
        // libc::sigemptyset(&raw mut mask);
        // libc::sigaddset(&raw mut mask, libc::SIGCHLD);
        // libc::sigprocmask(libc::SIG_UNBLOCK, &raw const mask, std::ptr::null_mut());
        libc::raise(libc::SIGCHLD);
    };

    assert_eq!(1, called.load(Ordering::Relaxed));
}

```

PR #758, opened 2025-09-18
Add and use a smaller webp version of logo

akx:
I noticed the logo in the GitHub README taking its sweet time to load even on a faster connection... well, no surprise since it's almost 3 megabytes.

This PR adds a smaller (512x512, the original being 1024x1024) WEBP version of the logo for use wherever print quality isn't needed.


j178:
Thank you!

PR #766, opened 2025-09-20
Add support for specifying `PREK_UV_SOURCE`

j178:


ulgens:
@j178 Is it possible not to install uv but use the existing one?

j178:
@ulgens Yes, prek reuses uv if it is installed: https://github.com/j178/prek/blob/eef84cf8c5fc7f96eb2b2398ab1747d56d678dc3/src/languages/python/uv.rs#L462

ulgens:
> Yes, prek reuses uv if it is installed:

Thank you, that's nice. Any chance to mention this in docs? From what I read, I was thinking it will always install a new one.

j178:
@ulgens Are you interested in contributing? This is an open source project and I’d be happy to accept PRs.

ulgens:
That's a fair response 🙂  I don't think I'll be available any time soon but I can add this to my todo list and revisit next week.

ulgens:
@j178 It seems you handled it before I could 🎉

PR #792, opened 2025-09-24
Deduplicate docs between README and MkDocs site

mondeja:
Hi,  

First of all, thanks for prek! I even started prototyping a rewrite of pre-commit myself but never had the time to finish it, so you’ve really saved me a lot of effort.  

Looking at the docs, it seems there are quite a few duplicated sections. I ran into the same issue a while back, which is why I created [`mkdocs-include-markdown-plugin`](https://github.com/mondeja/mkdocs-include-markdown-plugin). It has been stable for years now.  

This PR removes most of the duplication in the docs while keeping the same TOC structure you created (I assume that’s for SEO reasons). The hooks are needed to convert links like `https://prek.j178.dev/benchmark/` into `benchmark.md` so they also work when serving locally.

Cheers!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,8 @@
 /target
 
 .cache
+__pycache__/
+site/
 
 # Insta snapshots.
 *.pending-snap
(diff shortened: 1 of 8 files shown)


j178:
Thank you!

Issue #621, opened 2025-09-04
prek.ci


grst:
Hi, 

great to see that some limitations of pre-commit are finally being overcome! I was wondering if there's  any recommendation how to run prek on CI to 
 * check 
 * apply auto-fixes
 * autoupdate the pre-commit-config.yaml

I could do this using github actions, but due to GitHub's security restrictions this has two main limitations:
 * PRs from forks cannot commit autofixes because they don't get access to a token with write permissions
 * Checks are not re-run on autofix commits, as a precaution to avoid infinite loops

Pre-commit solves this by providing the pre-commit.ci service as a GitHub App. 
Are there plans to host a similar service for prek, or plans to provide such an app to self-host, or any other ideas? 

CC @flying-sheep

j178:
Thanks for the advice! 
A hosted service seems like overkill. It’s not just about time; it’s also a big money investment. Maybe a self-hosted GitHub App is more doable for me. I’ll think about it once I’ve wrapped up the main features.

LordFckHelmchen:
@grst Wouldn't a PAT work for the autoupdate workflow? This way you could trigger the other CI actions... (see https://github.com/orgs/community/discussions/57484#discussioncomment-6116484)

grst:
A PAT would solve the issue that checks are not triggered, but it wouldn't solve the issue with forks. 

Also, GitHub discourages the use of a PAT for such use-cases: 

> If you want to access GitHub resources on behalf of a user or in an organization, or you anticipate a long-lived integration, we recommend building a GitHub App.
>
> You can use personal access tokens for API testing or short-lived scripts. Since a personal access token is associated with a user, your automation could break if the user no longer has access to the resources you need. A GitHub App installed on an organization is not dependent on a user. Additionally, unlike a user, a GitHub App does not consume a GitHub **seat.**

https://docs.github.com/en/apps/creating-github-apps/about-creating-github-apps/deciding-when-to-build-a-github-app#choosing-between-a-github-app-or-a-personal-access-token

ulgens:
> check
> apply auto-fixes
> autoupdate the pre-commit-config.yaml

Personally, I never liked these 3 coming together as a hosted service.

> check

This is a CI task. I much prefer having that on the same platform as my existing CI. Offloading it to an external platform that I have no way of monitoring is not a positive value.

> apply auto-fixes

[...]

PR #800, opened 2025-09-26
Add a hint when hooks failed in CI

j178:
From https://github.com/j178/prek/issues/621#issuecomment-3299057336


potiuk (approved):
Nice! Thanks for being receptive :)

PR #801, opened 2025-09-26
Fix interpreter identification

mondeja:
Fixes interpreter identification. For example, a file without extension with the shebang `#!/usr/bin/env sh` is not being identified as a Shell script.

See https://github.com/mondeja/prek-interpreter-identify-bug repository and the next GHA run for reproduction: https://github.com/mondeja/prek-interpreter-identify-bug/actions/runs/18036744469/job/51325321188


j178:
Thank you!

PR #828, opened 2025-10-01
Support pass multiple files like `prek run --files a b c d`

### LINKED ISSUE #825: Support pass multiple files like `prek run --files a b c d`
Thanks for building this tool! I found a small difference in behavior between `prek` and `pre-commit`:

```bash
$ prek run --files a b c d
warning: the following selectors did not match any hooks or projects:
  - `b`
  - `c`
  - `d`
error: No hooks found after filtering with the given selectors
```

`pre-commit` handles this correctly:

```bash
$ pre-commit run --files a b c d
check for added large files..........................(no files to check)Skipped
check for merge conflicts............................(no files to check)Skipped
```

It looks like `pre-commit` treats all arguments after `--files` as file paths, while `prek` treats only the first argument as a file and interprets the remaining tokens as selectors.

This limitation may come from `clap` — by default clap expects repeated flags like `--files a --files b --files c --files d` to pass multiple files. You can however accept multiple space-separated values with clap by using something like `num_args = 1..` (see: https://docs.rs/clap/latest/clap/builder/struct.Arg.html#method.num_args).

Is the current behavior in `prek` intentional? If not, I’d be happy to open a PR to align `prek` with `pre-commit` (e.g. update the clap argument to accept multiple values).

SigureMo:
Fixes #825

### DIFF
--- a/src/cli/mod.rs
+++ b/src/cli/mod.rs
@@ -402,6 +402,7 @@ pub(crate) struct RunArgs {
     #[arg(
         long,
         conflicts_with_all = ["all_files", "from_ref", "to_ref"],
+        num_args = 1..,
         value_hint = ValueHint::AnyPath)
     ]
     pub(crate) files: Vec<String>,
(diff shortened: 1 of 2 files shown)


j178:
Thanks!

PR #831, opened 2025-10-01
Allow run `--files` with zero args

SigureMo:
Related to #828

Sorry to bother you! I just remembered that the `nargs='*'` used in `pre-commit` corresponds to clap's `num_args = 0..`, which allows passing an empty argument sequence.

In my use case I collect the list of changed files for each PR using `git diff --name-only --diff-filter=ACMR` (for very large repos like [PaddlePaddle/Paddle](https://github.com/PaddlePaddle/Paddle) it's impractical to run full checks on every PR), and then pass that list to `prek run --files`. If a PR only deletes files, an empty list may be passed to `prek run --files`, which would cause an error.

Sorry for the earlier oversight — I should have noticed this from your reply.

### DIFF
--- a/src/cli/mod.rs
+++ b/src/cli/mod.rs
@@ -402,7 +402,7 @@ pub(crate) struct RunArgs {
     #[arg(
         long,
         conflicts_with_all = ["all_files", "from_ref", "to_ref"],
-        num_args = 1..,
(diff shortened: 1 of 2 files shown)


j178:
Thanks!

PR #852, opened 2025-10-08
Add a Dockerfile

j178:


ulgens:
What is the goal with this Dockerfile? What is the expected / planned usage?

j178:
It’s for people who need to build a Docker image for prek and I’m planning to add build and publish the Docker image as part of the release process.

PR #860, opened 2025-10-09
Ignore config error if the directory is skipped

j178:


m-aciek:
It doesn't seem to work, if I don't miss anything 🤔 

```
% prek --version
prek 0.2.8 (e924c4ba5 2025-10-14)
% prek install --skip taa
prek installed at `.git/hooks/pre-commit`
% git commit
error: Failed to parse `taa/.pre-commit-config.yaml`
  caused by: invalid type: string "../javelin/.pre-commit-config.yaml", expected struct Config
```

hook script:
```
#!/usr/bin/env bash
# File generated by prek: https://github.com/j178/prek
# ID: 182c10f181da4464a3eec51b83331688

# Unset GIT_DIR to avoid git taking current directory as the git dir.
# See:
# https://github.com/pre-commit/pre-commit/issues/2295
# https://www.spinics.net/lists/git/msg374197.html
unset GIT_DIR

ARGS=(hook-impl --skip=taa --hook-type=pre-commit --script-version=3)

HERE="$(cd "$(dirname "$0")" && pwd)"
ARGS+=(--hook-dir "$HERE" -- "$@")
PREK="/Users/maciej.olko/.local/bin/prek"

# Check if the full path to prek is executable, otherwise fallback to PATH
if [ ! -x "$PREK" ]; then
    PREK="prek"
fi

exec "$PREK" "${ARGS[@]}"
```

j178:
@m-aciek You need to use `prek install --skip taa/` - the trailing slash is important because it makes it a path selector. Without the slash, it gets recognized as a hook id selector instead.

https://prek.j178.dev/workspace/#selector-syntax

m-aciek:
Thank you! Yes, I found it in parallel. 👍 🚀 thank you for fixing and releasing this

PR #867, opened 2025-10-11
Fix autoupdate `rev` rendering for "float-like" version numbers

### LINKED ISSUE #866: `autoupdate` can break quoting for float-like version numbers
### Summary

The following config is broken by `autoupdate`:

```yaml
repos:
- repo: https://github.com/mgedmin/check-manifest
  rev: "0.49"
  hooks:
  - id: check-manifest
    additional_dependencies: [cmake, ninja]
```

After `autoupdate` this becomes

```yaml
repos:
- repo: https://github.com/mgedmin/check-manifest
  rev: "'0.50'"
  hooks:
  - id: check-manifest
    additional_dependencies: [cmake, ninja]
```

`prek` correctly discovers that the next version would be `0.50`, so the error seems to be in the rendering.

```console
$ prek autoupdate
[https://github.com/mgedmin/check-manifest] updating 0.49 -> 0.50
```

### Platform

macos15

### Version

prek 0.2.5 (33864f24b 2025-10-10)

### .pre-commit-config.yaml

_No response_

### Log file

_No response_

bbannier:
A version number could look like a float, e.g., 0.49 is a valid version number. Serde YAML serialization already renders such version numbers in quotes so types match the expectation. When doing an autoupdate we were explicitly adding more quotes if they were present in the input. This could have caused incorrect quoting if existing quoting clashed with serde's quoting style, e.g., if `"` where used but serde defaulted to quoting with `'`.

This patch removes any extra quoting here and fully defers to serde.

Closes #866.

### DIFF
--- a/src/cli/auto_update.rs
+++ b/src/cli/auto_update.rs
@@ -397,14 +397,6 @@ async fn write_new_config(path: &Path, revisions: &[Option<Revision>]) -> Result
             .expect("Invalid regex")
             .expect("Failed to capture revision line");
 
-        // TODO: preserve the quote style
-        // Naively add the original quotes
-        let new_rev = if !caps[3].is_empty() && !new_rev.contains(&caps[3]) {
-            format!("{}{}{}", &caps[3], new_rev.trim(), &caps[3])
-        } else {
-            new_rev.trim().to_string()
-        };
-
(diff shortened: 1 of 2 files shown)


j178:
Thanks, I think this makes sense.

Issue #880, opened 2025-10-14
[perf] Port more pre-commit hooks


lmmx:
I don't see a checklist tracking issue for the remaining pre-commit hooks, but some like check-toml have been implemented and gone from sub-second to instant, so this would likely contribute a further speedup on many existing pre-commit configs.

- Docs: [Currently implemented hooks](https://prek.j178.dev/builtin/?h=check+toml#httpsgithubcompre-commitpre-commit-hooks)
  - Source: [docs/builtin.md](https://github.com/j178/prek/blob/1205b6ddb62e3f0fc85f0f134d427bb2dcfcfd80/docs/builtin.md?plain=1#L31)

It looks like the specific site of the fast path redirection for these hooks is [here](https://github.com/j178/prek/blob/1205b6ddb62e3f0fc85f0f134d427bb2dcfcfd80/src/builtin/pre_commit_hooks/mod.rs#L32-L45) in `src/builtin/pre_commit_hooks/mod.rs`

To support this, I would compare the list in the docs against [this repo](https://github.com/pre-commit/pre-commit-hooks), specifically in [this hooks YAML](https://github.com/pre-commit/pre-commit-hooks/blob/main/.pre-commit-hooks.yaml)

## pre-commit-hooks

- [x] check-added-large-files
- [ ] check-ast
- [ ] check-byte-order-marker
- [ ] check-builtin-literals
- [ ] check-case-conflict
- [ ] check-docstring-first
- [ ] check-executables-have-shebangs
- [ ] check-illegal-windows-names
- [x] check-json
- [ ] check-shebang-scripts-are-executable
- [ ] pretty-format-json
- [ ] check-merge-conflict
- [ ] check-symlinks
- [x] check-toml
- [ ] check-vcs-permalinks
- [ ] check-xml
- [x] check-yaml
- [ ] debug-statements
- [ ] destroyed-symlinks
- [ ] detect-aws-credentials
- [ ] detect-private-key
- [ ] double-quote-string-fixer
- [x] end-of-file-fixer
- [ ] file-contents-sorter
- [x] fix-byte-order-marker
- [ ] fix-encoding-pragma
- [ ] forbid-new-submodules
- [ ] forbid-submodules
- [x] mixed-line-ending
- [ ] name-tests-test
- [ ] no-commit-to-branch
- [ ] requirements-txt-fixer
- [ ] sort-simple-yaml
- [x] trailing-whitespace

j178:
[...]

PR #884, opened 2025-10-14
feat(check-ast): implement builtin hook

lmmx:
## Description

Following discussion in #880, noticing that the `check-ast` built in hook is [quite popular](https://github.com/j178/prek/issues/880#issuecomment-3402811333) (2.7k, mid-tier), I thought I'd see if I could implement it using the `ruff_python_parser` used internally by ruff, since we are not trying to ship to crates.io (therefore can use git dependencies).

- [x] The extra dependency increased compile time (after `cargo clean` each time) +4%, from 21.9s to 22.8s, which is hopefully reasonable.
- [x] Included unit tests and snapshot tests

[Binary Size Change](https://github.com/j178/prek/actions/runs/18629115999/job/53111278636#step:9:356)
+8.75% (.text: 16.0 MiB → 17.4 MiB)

## Demo

Taking an example repo which uses it, [chatarena](https://github.com/Farama-Foundation/chatarena/blob/main/.pre-commit-config.yaml)

```sh
gh repo clone Farama-Foundation/chatarena
```

After commenting out other hooks and running the suites with regular pre-commit/prek:

```
runpc --verbose
```

```
check python ast.........................................................Passed
- hook id: check-ast
- duration: 0.06s
```

```sh
runpk --verbose
```

```
check python ast.........................................................Passed
- hook id: check-ast
- duration: 0.09s
```

prek is +50% slower (+0.03s) than pre-commit (0.06s)

After installing `prek` from this feature branch with `cargo install --path .` the speed goes to 0.00s :tada: 

```
louis 🌟 ~/tmp/chatarena $ pre-commit run -a --verbose
check python ast.........................................................Passed
- hook id: check-ast
- duration: 0.05s
louis 🌟 ~/tmp/chatarena $ prek run -a --verbose
check python ast.........................................................Passed
- hook id: check-ast
- duration: 0.00s
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -193,6 +193,36 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "1505bd5d3d116872e7271a6d4e16d81d0c8570876c8de68093a09ac269d8aac0"
 
 [[package]]
+name = "attribute-derive"
+version = "0.10.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "05832cdddc8f2650cc2cc187cc2e952b8c133a48eb055f35211f61ee81502d77"
+dependencies = [
+ "attribute-derive-macro",
+ "derive-where",
+ "manyhow",
+ "proc-macro2",
+ "quote",
+ "syn",
+]
+
+[[package]]
+name = "attribute-derive-macro"
+version = "0.10.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0a7cdbbd4bd005c5d3e2e9c885e6fa575db4f4a3572335b974d8db853b6beb61"
+dependencies = [
+ "collection_literals",
+ "interpolator",
+ "manyhow",
+ "proc-macro-utils",
+ "proc-macro2",
+ "quote",
+ "quote-use",
+ "syn",
+]
+
+[[package]]
 name = "autocfg"
 version = "1.5.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


j178:
Thank you!
I’m a bit concerned about how much the binary size might increase after pulling in ruff_python_ast and ruff_python_parser. Could you compare the binary sizes when building with the release profile?

lmmx:
Sure thing, I just checked the master branch release build size in #894 as:

- master branch: 11MB (10552B)

For this one (after `cargo clean` and rebuild with `--release`) I get `du target/release/prek` as:

- check-ast branch: 12MB (11908B)

So +12.8% (+1356B) bigger binary, but it'd be reusable across other hooks, so I think it's worth it!

- #886
- #890

Yep, they match:

debug-statements branch

```sh
louis 🌟 ~/dev/prek $ du -h target/release/prek; echo; du target/release/prek
12M     target/release/prek

11920   target/release/prek
```

lmmx:
> I’m a bit concerned about how much the binary size might increase after pulling in ruff_python_ast and ruff_python_parser. Could you compare the binary sizes when building with the release profile?

[New cargo bloat CI check](https://github.com/j178/prek/pull/927#issuecomment-3419562766) shows it as

> +8.75% (.text: 16.0 MiB → 17.4 MiB)

I presume since the build that ships is done on CI, this will be the more accurate one

j178:
Thanks for putting in the effort on this! I've decided not to go with this one. Here's my thinking:

- It's not that widely used
- It brings in a pretty big dependency just for Python support
- It adds a git dependency, and since I want prek to be publishable on crates.io, that's gonna be a problem

Feel free to reopen or discuss more if you want.

lmmx:
No it was fun to work with the ruff AST parser crate for the first time, no harm done!

PR #885, opened 2025-10-14
feat(check-merge-conflict): implement builtin hook

lmmx:
## Description

Following discussion in #880 I decided to speed up the next most commonly used pre-commit hook (7.8k repos on grep.app as listed [here](https://github.com/j178/prek/issues/880#issuecomment-3402811333)).

This checks for merge conflict markers, and [in the original](https://github.com/pre-commit/pre-commit-hooks/blob/main/pre_commit_hooks/check_merge_conflict.py) takes a `--assume-in-merge` flag and has some `CONFLICT_PATTERNS`. This PR does the same but in Rust, with unit tests and snapshot tests like the other hooks.

- [x] No extra dependencies
- [x] Included unit tests and snapshot tests

## Demo

Installing it locally and running on the top hit for this hook on grep.app, [PaddleSpeech](https://github.com/PaddlePaddle/PaddleSpeech):

```sh
gh repo clone PaddlePaddle/PaddleSpeech
```

prek runs the `check-merge-conflict` hook 40% slower than `pre-commit` (.07s -> .10s)

```
louis 🌟 ~/tmp/PaddleSpeech $ pre-commit run -a --verbose
check for merge conflicts................................................Passed
- hook id: check-merge-conflict
- duration: 0.07s
check json...............................................................Passed
- hook id: check-json
- duration: 0.03s
louis 🌟 ~/tmp/PaddleSpeech $ prek run -a --verbose
check for merge conflicts................................................Passed
- hook id: check-merge-conflict
- duration: 0.10s
check json...............................................................Passed
- hook id: check-json
- duration: 0.00s
```

But once ported to Rust it becomes instant :tada: 

```
louis 🌟 ~/tmp/PaddleSpeech $ cargo install --path ~/dev/prek 2>/dev/null
louis 🌟 ~/tmp/PaddleSpeech $ prek run -a --verbose
check for merge conflicts................................................Passed
- hook id: check-merge-conflict
- duration: 0.00s
check json...............................................................Passed
- hook id: check-json
- duration: 0.00s
```

### DIFF
--- a/src/builtin/pre_commit_hooks/check_merge_conflict.rs
+++ b/src/builtin/pre_commit_hooks/check_merge_conflict.rs
@@ -4,6 +4,7 @@ use anyhow::Result;
 use clap::Parser;
 use futures::StreamExt;
 
+use crate::git::get_git_dir;
 use crate::hook::Hook;
 use crate::run::CONCURRENCY;
 
@@ -31,7 +32,7 @@ pub(crate) async fn check_merge_conflict(
     let args = Args::try_parse_from(hook.entry.resolve(None)?.iter().chain(&hook.args))?;
 
     // Check if we're in a merge state or assuming merge
-    if !args.assume_in_merge && !is_in_merge(hook.work_dir()).await? {
+    if !args.assume_in_merge && !is_in_merge().await? {
         return Ok((0, Vec::new()));
     }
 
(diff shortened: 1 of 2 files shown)


j178:
Thank you! You're so efficient! I'll review and provide feedback ASAP - thanks for all your contributions!

lmmx:
You’re most welcome, I realised after writing one that it’d be easy to do another after getting the idea of what needed changing 😄

src/builtin/pre_commit_hooks/check_merge_conflict.rs
```diff
@@ -0,0 +1,297 @@
...
+}
+
+async fn is_in_merge(work_dir: &Path) -> Result<bool> {
+    // Get the git directory
+    let git_dir_output = tokio::process::Command::new("git")
+        .arg("rev-parse")
+        .arg("--git-dir")
+        .current_dir(work_dir)
+        .output()
+        .await?;
```
j178: We can reuse `git::get_git_dir()` here

lmmx: Patched in 2fc2e9d

src/builtin/pre_commit_hooks/check_merge_conflict.rs
```diff
@@ -0,0 +1,280 @@
...
+    // Check if any of the merge state files exist
+    let merge_head_exists = git_dir.join("MERGE_HEAD").exists();
+    let rebase_apply_exists = git_dir.join("rebase-apply").exists();
+    let rebase_merge_exists = git_dir.join("rebase-merge").exists();
+
+    Ok(merge_head_exists || rebase_apply_exists || rebase_merge_exists)
+}
+
+async fn check_file(file_base: &Path, filename: &Path) -> Result<(i32, Vec<u8>)> {
+    let content = fs_err::tokio::read(file_base.join(filename)).await?;
```
j178, line 74: I'm thinking it might be better to read the file line by line (like the original implementation did) rather than loading the whole thing into memory. What do you think?

lmmx: No I'd keep reading the entire file, we'll typically be calling the hook with many files (and if there aren't many files it'll probably be instantaneous anyway). I'd expect reading line by line might itself add overhead

If you wanted to optimise it I'd reach for memchr (I didn't notice this is already a dependency until I'd already sent some PRs in but that's very useful for fast file ops e.g. it can find the line boundaries for us)

I'll see if I can benchmark a version using that instead...

lmmx: So it actually runs in 0.00s on the Airflow repo so I just put a dozen of the same hooks in to run it multiple times... Hopefully that's a legitimate way to check perf (maybe we should have a `--debug` mode that adds more decimal places to the duration time print out).

Anyway, this feature branch currently runs the 32 hooks in 0.32 seconds so 0.01s per run by the looks of it.

...and rewritten with memchr I can't detect a difference in either run duration or resident set size (RAM load), I don't think it's worth rewriting

lmmx: I agree though line by line is simpler :+1:

j178:
Thank you!

lmmx:
Too speedy haha! :rofl: My last commit didn't get included, sorry shouldn't have clicked resolved without pushing (extra bit is in #907)

Issue #872, opened 2025-10-13
Slower to run some hooks (debug-statements, ruff) than pre-commit


lmmx:
### Summary

I’ve just taken prek for a spin and while the hook installation was faster (maybe 3-4x!), meaning overall run the first time (after wiping the cache dirs for level playing field) was faster, subsequent runs were slower (by about 0.5s - 1s) because somehow prek is slower to execute the hooks than pre-commit.

I presume this is worth a bug report given speed is prek’s primary goal, and hooks are likely going to be executed more often than they are installed (at least locally)

### Platform

Linux 5.15.0-125-generic x86_64 GNU/Linux

### Version

prek 0.2.5

### .pre-commit-config.yaml

<details><summary>Click to show</summary>

```
ci:
  autofix_commit_msg: "chore(pre-commit): autofix run"
  autoupdate_commit_msg: "chore(pre-commit): autoupdate hooks"

default_install_hook_types:
  - pre-commit
  - commit-msg

repos:
  - repo: https://github.com/compilerla/conventional-pre-commit
    rev: v4.2.0
    hooks:
      - id: conventional-pre-commit
        stages: [commit-msg]
        # args: [build, chore, ci, docs, feat, fix, perf, refactor, revert, style, test]
  - repo: https://github.com/pre-commit/pre-commit-hooks
    rev: v6.0.0
    hooks:
      - id: check-toml
      - id: check-yaml
        exclude: ^mkdocs.yml
      - id: debug-statements
      - id: end-of-file-fixer
      - id: name-tests-test
      - id: trailing-whitespace
  - repo: https://github.com/pappasam/toml-sort
    rev: v0.24.3
    hooks:
      - id: toml-sort-fix
  - repo: https://github.com/asottile/add-trailing-comma
    rev: v3.2.0
    hooks:
      - id: add-trailing-comma
  - repo: https://github.com/asottile/pyupgrade
    rev: v3.20.0
    hooks:
      - id: pyupgrade
        args: [--py310-plus]
  - repo: https://github.com/astral-sh/ruff-pre-commit
    rev: v0.13.3
    hooks:
      # Run the linter
      - id: ruff
        args: [--fix]
      # Run the formatter
      - id: ruff-format
```

</details>

### Log file

<details><summary>Click to show</summary>

```
2025-10-13T09:32:47.825851Z DEBUG prek: 0.2.5 (/home/louis/.local/bin/prek)
[...]

PR #886, opened 2025-10-14
feat(debug-statements): implement builtin hook

lmmx:
## Description

Following discussion in #880 and initial bug report in #872 (where found prek slower than pre-commit for the `debug-statements` hook in particular, which can be fixed by giving it a Rust port), this PR contributes such a "fast path" hook.

This is a popular hook with 5,000+ repos using it [on grep.app](https://github.com/j178/prek/issues/880#issuecomment-3402811333)

Like #884 this one uses the `ruff_python_parser` crate from ruff to parse the AST and then its closely related crate `ruff_python_ast` to access nodes.

- [x] The extra crate increased compilation time by 3%, from 21.7s to 22.4s (and I would expect this to be negligible difference to the change introduced by `ruff_python_parser` alone in the `check-ast` hook PR in #884)
- [x] Unit test and snapshot test coverage added

[Binary Size Change](https://github.com/j178/prek/actions/runs/18629115264/job/53111276775#step:9:356)
+8.75% (.text: 16.0 MiB → 17.4 MiB)

## Demo

Taking the top hit on grep.app, [wandb/wandb](https://github.com/wandb/wandb/blob/main/.pre-commit-config.yaml)

```sh
gh repo clone wandb/wandb
```

and running the latest release of prek and pre-commit, the `debug-statements` hook runs 26% slower than pre-commit's:

```sh
louis 🌟 ~/tmp/wandb $ pre-commit run -a --verbose
check-toml...............................................................Passed
- hook id: check-toml
- duration: 0.04s
debug-statements.........................................................Passed
- hook id: debug-statements
- duration: 0.19s
louis 🌟 ~/tmp/wandb $ prek run -a --verbose
check-toml...............................................................Passed
- hook id: check-toml
- duration: 0.00s
debug-statements.........................................................Passed
- hook id: debug-statements
- duration: 0.24s
```

After installing this local feature branch, it runs 2x as fast :tada:

```sh
louis 🌟 ~/tmp/wandb $ prek run -a --verbose
check-toml...............................................................Passed
- hook id: check-toml
- duration: 0.00s
debug-statements.........................................................Passed
- hook id: debug-statements
- duration: 0.10s
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -193,6 +193,36 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "1505bd5d3d116872e7271a6d4e16d81d0c8570876c8de68093a09ac269d8aac0"
 
 [[package]]
+name = "attribute-derive"
+version = "0.10.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "05832cdddc8f2650cc2cc187cc2e952b8c133a48eb055f35211f61ee81502d77"
+dependencies = [
+ "attribute-derive-macro",
+ "derive-where",
+ "manyhow",
+ "proc-macro2",
+ "quote",
+ "syn",
+]
+
+[[package]]
+name = "attribute-derive-macro"
+version = "0.10.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


j178:
Similar to https://github.com/j178/prek/pull/884#issuecomment-3552391372, I’m gonna close this one too. Thanks for your effort though!

PR #888, opened 2025-10-14
Implement `check-case-conflict` as builtin hook

lmmx:
## Description

Implementing the `check-case-conflict` hook which ensures you don't check in files that would be the same on case-insensitive file systems (I think this is a Windows thing).

This is a widely used hook, 4.8k repos on grep.app.

- [x] No new dependencies.
- [x] It needed one new `git` function to mirror [the equivalent one](https://github.com/pre-commit/pre-commit-hooks/blob/9ba250d7b35601b5ad49f661ce18783127d0b41c/pre_commit_hooks/util.py#L11-L13) in the Python original implementation.
- [x] Unit tests and snapshot tests included

[Binary Size Change](https://github.com/j178/prek/actions/runs/18629114297/job/53111274368#step:9:312)
+0.63% (.text: 16.0 MiB → 16.1 MiB)

## Demo

The most popular repo with this hook is [jupyter/notebook](https://github.com/jupyter/notebook/blob/main/.pre-commit-config.yaml)

The hook takes 0.08s to run on pre-commit and 0.12s on prek (50% slower)

```sh
louis 🌟 ~/tmp/notebook $ runpc --verbose
check for case conflicts.................................................Passed
- hook id: check-case-conflict
- duration: 0.08s
check toml...............................................................Passed
- hook id: check-toml
- duration: 0.03s
check yaml...............................................................Passed
- hook id: check-yaml
- duration: 0.04s
```

```sh
louis 🌟 ~/tmp/notebook $ runpk --verbose
real    0m0.339s
user    0m1.255s
sys     0m0.292s
check for case conflicts.................................................Passed
- hook id: check-case-conflict
- duration: 0.12s
check toml...............................................................Passed
- hook id: check-toml
- duration: 0.00s
check yaml...............................................................Passed
- hook id: check-yaml
- duration: 0.00s
```

After installing from this local feature branch version, it runs instantly :tada: 

```sh
louis 🌟 ~/tmp/notebook $ time runpk --verbose
check for case conflicts.................................................Passed
- hook id: check-case-conflict
- duration: 0.00s
```

### DIFF
--- a/crates/prek/src/hooks/pre_commit_hooks/check_case_conflict.rs
+++ b/crates/prek/src/hooks/pre_commit_hooks/check_case_conflict.rs
@@ -1,5 +1,5 @@
 use std::collections::hash_map::Entry;
-use std::path::{Path, PathBuf};
+use std::path::Path;
 
 use anyhow::Result;
 use rustc_hash::FxHashMap;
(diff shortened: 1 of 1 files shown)


lmmx:
Ha! The snapshots have broken on Windows and macOS because those filesystems are indeed case insensitive by default so the files are never produced so there is no conflict, as the filesystem prevents you from creating those files in the same directory so the conflicts never exist in the first place!

Screenshot from 2025-10-14 23-25-14

In which case: those tests should be skipped on those platforms

**edit** rather than skip those tests depending on platform, I returned early so as to confirm the assumption still holds (it did), now all the CI tests pass

lemonyte:
This would be the last of the popular (>2k users) and not python-specific hooks that's missing. I'm one of the users who relies on this hook, so it would be awesome to see it merged!
Thanks @lmmx for your work and @j178 for the project!

j178:
Thanks @lmmx!

PR #892, opened 2025-10-15
[draft] feat(implement-prek-gc): implement gc command

feliblo:
I was bored so I gave it a try. 

Would love your thoughts and differences between what you've implemented yourself so far @j178. 

Don't wish to step on anyone's toes. Happy to close the requests if your implementation is further/better or a different direction than mine. I based it completely on the [pre-commit implementation](https://github.com/pre-commit/pre-commit/blob/main/pre_commit/commands/gc.py).

**Personal Todos**:
- [ ] Add documentation for the `gc` command
- [ ] Review tests


tests/gc.rs
```diff
@@ -0,0 +1,1016 @@
...
+
+    // Create pyproject.toml for workspace detection
+    work_dir
+        .child("pyproject.toml")
+        .write_str(indoc::indoc! {r#"
+        [tool.prek.workspace]
+        members = ["project1", "project2"]
+    "#})?;
+
+    cmd_snapshot!(context.filters(), context.command().arg("gc").env("PREK_HOME", &*home).current_dir(work_dir), @r"
```
Felix-Blom: note: seems weird to me to gc here, since it is still in usage of other projects. What's intended?

tests/gc.rs
```diff
@@ -0,0 +1,1016 @@
...
+    success: false
+    exit_code: 101
+    ----- stdout -----
+
+    ----- stderr -----
+    warning: Both `[TEMP_DIR]/.pre-commit-config.yaml` and `[TEMP_DIR]/.pre-commit-config.yml` exist, using `[TEMP_DIR]/.pre-commit-config.yaml` only
+
+    thread 'main' panicked at src/workspace.rs:688:9:
+    At least one project should be found
```
feliblo, line 356: Different note: shouldn't this general error message contain more information than: At least one project should be found? In this case it exists, but is malformed.

e.g.: 'At least one project should be found. If a project file exists it may be malformed.' or
'At least one project should be found. The project files might not exist or be malformed.' 

_I could handle it in this/different PR_

j178:
Thank you for working on this!

I’ve been dragging my feet on implementing this feature and haven’t made much progress.

But looking at how the original pre-commit handles gc, I don’t think `prek gc` should just keep the current project’s hooks. We need to track all projects when users run `prek run` - in my implementation, it’s a set of configuration files path that I persist in `~/.cache/prek/config-tracking.json`. Every time `prek run` runs in a new project, it gets added to the file.

When running `prek gc`, we’d load all the configs, find all the associated repos, hooks, and tools, then clean up any unreachable files.

tests/gc.rs
j178: Yeah, this is currently a debug_assert! because I figured it was a logic error https://github.com/j178/prek/blob/bb9cd2d30ac96b5a323f0094b04d49598dca2fc1/src/workspace.rs#L688

But I’m thinking we should actually turn it into a proper user-facing error instead.

j178: https://github.com/j178/prek/issues/911

PR #898, opened 2025-10-15
fix: system language version

### LINKED ISSUE #698: "Invalid `language_version` value: `system`" encountered with `npm`-based local hooks
Hi! Thank you for developing `prek` – I've been enjoying using it quite a lot recently, and it's super fast! 💜

I recently came across an issue where I've set `default_language_version` for `node` as `system` – which should be accepted as per the code, but it `prek` complains with the following:

```bash
error: Hook `prettier` is invalid
  caused by: Invalid `language_version` value: `system`
```

Here is a minimal `pre-commit-config.yaml` file that can reproduce the issue:

```yaml
default_language_version:
  node: system

repos:
  - repo: local
    hooks:
      - id: prettier
        name: prettier
        entry: 'npm run prettier'
        language: node
        types_or: [json, ts, tsx, javascript, jsx, css]
        exclude: \.ipynb$
      - id: eslint
        name: eslint
        entry: 'npm run eslint'
        language: node
        types_or: [ts, tsx, javascript, jsx]
      - id: stylelint
        name: stylelint
        entry: 'npm run stylelint'
        language: node
        types: [css]
```

The same error occurs with `eslint` and `stylelint`, i.e., if I comment out `prettier` – so I assume it will occur with all `npm`-based hooks:

```bash
error: Hook `eslint` is invalid
  caused by: Invalid `language_version` value: `system`
```

```bash
error: Hook `stylelint` is invalid
  caused by: Invalid `language_version` value: `system`
```

On the other hand, `pre-commit` does not raise an error with my configuration file.

My system specifications are as follows:
- macOS M-series machine with Sequoia 15.6.1
- `prek` has been installed with `uv` globally via the `uv tool install prek` command
- The `prek --version` command reports `prek 0.1.6 (2092da67d 2025-09-03)`, i.e., the latest available stable release at the time of writing.

<hr>

Thank you for your time!

lmmx:
## Description

- Closes #698 

Simple fix to the bug reported, does what pre-commit does now

Screenshot from 2025-10-15 18-56-57

I don't really get what "language_version: system" means but my impression is we are doing the same as pre-commit

### DIFF
--- a/tests/languages/system.rs
+++ b/tests/languages/system.rs
@@ -1,0 +1,61 @@
+use crate::common::{TestContext, cmd_snapshot};
+
+#[test]
+fn system_language_version() {
+    let context = TestContext::new();
+    context.init_project();
+    context.write_pre_commit_config(indoc::indoc! {r"
+        default_language_version:
+          node: system
+        repos:
+          - repo: local
+            hooks:
+              - id: node-version
+                name: node-version
+                language: node
+                entry: node --version
+                always_run: true
+                pass_filenames: false
+    "});
+    context.git_add(".");
+
+    cmd_snapshot!(context.filters(), context.run(), @r"
+    success: true
+    exit_code: 0
+    ----- stdout -----
+    node-version.............................................................Passed
+
+    ----- stderr -----
+    ");
+}
+
+#[test]
+fn system_hooks() {
+    let context = TestContext::new();
+    context.init_project();
+    context.write_pre_commit_config(indoc::indoc! {r"
+        repos:
+          - repo: local
+            hooks:
+              - id: echo
+                name: echo
+                language: system
+                entry: echo hello world
+                always_run: true
+                pass_filenames: false
+                verbose: true
+    "});
+    context.git_add(".");
+
+    cmd_snapshot!(context.filters(), context.run(), @r#"
+    success: true
+    exit_code: 0
+    ----- stdout -----
+    echo.....................................................................Passed
+    - hook id: echo
+    - duration: [TIME]
+      hello world
+
+    ----- stderr -----
+    "#);
+}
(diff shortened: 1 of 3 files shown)


j178:
I think `language_version: system` has an implication that it should just use whatever toolchain is already installed on the system, without trying to download a new one. This PR makes `system` work the same as default, which does work, but we’re losing that “don’t install anything new” behavior that `system` was supposed to have.

tests/languages/system.rs
```diff
@@ -0,0 +1,61 @@
...
+fn system_hooks() {
+    let context = TestContext::new();
+    context.init_project();
+    context.write_pre_commit_config(indoc::indoc! {r"
+        repos:
+          - repo: local
+            hooks:
+              - id: echo
+                name: echo
+                language: system
```
j178, line 42: I think this test isn't relevant - `language: system` is a different thing from `language_version: system`.

lmmx: ah you’re right it doesn’t repro — sorry I’d got it in my head that `language: system` implied `language_version: system` but trying it out I see that’s not true and the original (reported in #698) I’d been running had

```yaml
default_language_version:
  node: system
```

at the top

lmmx:
> I think `language_version: system` has an implication that it should just use whatever toolchain is already installed on the system, without trying to download a new one. This PR makes `system` work the same as default, which does work, but we’re losing that “don’t install anything new” behavior that `system` was supposed to have.

I see what you mean, there is meant to be logic per-language like `NodeRequest` for `language: node`. We probably instead want to implement a `LanguageRequest::System` variant and subsequent per-language variants

- [ ] node: `NodeRequest`
- [ ] python: `PythonRequest`
- [ ] go: `GoRequest`

j178:
Closing in favor of https://github.com/j178/prek/pull/949, thanks for your work!

Issue #904, opened 2025-10-16
Add pre-commit copyright to LICENSE


lmmx:
Hi! Just got a note from the pre-commit maintainer that this needs to include their copyright line in the LICENSE file to comply with the MIT license terms.

To fix, it just needs this line above the existing copyright in LICENSE:

```
Copyright (c) 2014 pre-commit dev team: Anthony Sottile, Ken Struys
```

- via https://github.com/pre-commit/pre-commit-hooks/issues/1201#issuecomment-3410938674

lmmx:
Nope, confirmed it was just a misunderstanding and the current licenses are OK (for both pre-commit and pre-commit-hooks repositories) in reply [here](https://github.com/pre-commit/pre-commit-hooks/issues/1201#issuecomment-3411500180)

PR #905, opened 2025-10-16
fix: Add pre-commit copyright line to LICENSE

lmmx:
- via #904

Include the original pre-commit copyright to comply with MIT license requirements for derivative works.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,5 +1,6 @@
 MIT License
 
+Copyright (c) 2014 pre-commit dev team: Anthony Sottile, Ken Struys
 Copyright (c) 2024 j178
 
 Permission is hereby granted, free of charge, to any person obtaining a copy


j178:
It’s included in https://github.com/j178/prek/blob/master/licenses/LICENSE.pre-commit.txt. I don’t know why Anthony keeps insisting that prek doesn’t include his license.

Also check out https://github.com/j178/prek/pull/74 - I don’t think Anthony will be satisfied with this approach either.

lmmx:
> It’s included in [`master`/licenses/LICENSE.pre-commit.txt](https://github.com/j178/prek/blob/master/licenses/LICENSE.pre-commit.txt?rgh-link-date=2025-10-16T14%3A33%3A47Z).

Oh! I wasn't aware you'd discussed it before, maybe he didn't realise you put it in.

> I don’t know why Anthony keeps insisting that prek doesn’t include his license.

Maybe just looked at the LICENSE file, I can mention that it's already in a licenses subdir and clarify if he'd prefer it to be in the single LICENSE. It is more typical to just add the Copyright line as far as I'm aware, you often see multiple Copyright lines where successive authors have built on each others' work.

> Also check out #74 - I don’t think Anthony will be satisfied with this approach either.

Honestly it's a shame that there was such hostility in that thread, it started out looking like it was going to be a fusion.

I'll see if I can get any clarity about the LICENSE situation

lmmx:
Nope, turns out it was just a misunderstanding and prek has the required licenses for pre-commit and pre-commit-hooks (since they're the same), so no license change needed.

Since that's cleared up now I will close this PR

j178:
Thank you!❤️

PR #907, opened 2025-10-16
refactor(merge-conflict): linewise reading

lmmx:
I sent this commit just at  the same moment you merged I think! :sweat_smile:

Just addresses the final review comment in #885


lmmx:
Ah never mind, sorry responded to that review while in the middle of something. Let’s keep it as is

PR #908, opened 2025-10-16
Add `--fail-fast` CLI flag to stop after first hook failure

### LINKED ISSUE #896: Add `prek run --fail-fast ` flag
See https://github.com/pre-commit/pre-commit/pull/3528

AdityasWorks:
##  Summary

Implements the `--fail-fast` CLI flag for the `prek run` command to stop hook execution immediately after the first failure, providing faster feedback during development.

#### CLI: 
` prek run --fail-fast `
#### Config file: 
` fail_fast = true `

- CLI flag OR config setting will enable fail-fast mode
- CLI flag takes precedence (enables feature, **cannot disable config setting**)

### The priority system works like this:

- If you use `--fail-fast` on the command line -> fail-fast is enabled
- If you have `fail_fast: true` in your config file -> fail-fast
- If neither is set then fail-fast is disabled (default behaviour)

### Here's what was tested:

- The code compiles without errors
- All existing hook implementation tests pass (8 out of 8)
- The existing fail-fast test passes
- Overall test results: 37 out of 38 tests pass

**One test  fails because it checks the help text output, which now includes our new --fail-fast flag.**

## Related Issue
Fixes #896

### DIFF
--- a/tests/run.rs
+++ b/tests/run.rs
@@ -766,7 +766,7 @@ fn fail_fast() {
     "#);
 }
 
-/// Test --fail-fast CLI flag stops execution after first failure. 
+/// Test --fail-fast CLI flag stops execution after first failure.
 #[test]
 fn fail_fast_cli_flag() {
     let context = TestContext::new();


j178:
Thanks! Can you add a test in tests/run.rs for this flag too?

j178 (approved):
Thanks!

PR #914, opened 2025-10-17
Add collision detection for hook env directories

### LINKED ISSUE #325: More robust hook direcotry generation
https://github.com/j178/prefligit/blob/9c64f463ff457c61568f40e35639768ba62ee8c7/src/hook.rs#L683-L690

kenwoodjw:
Closed #325


j178:
Thanks for working on this! I switched to using tempfile for more reliable detection and to create the directory safely. This helps avoid [TOCTOU](https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use) issues.

PR #924, opened 2025-10-18
Implement `check-executables-have-shebangs` as builtin-hook

feliblo:
## Description

Following the information provided in [880](https://github.com/j178/prek/issues/880). This is an implementation of the `check-executables-have-shebangs` hook.

The [check-executables-have-shebangs](https://grep.app/search?f.path.pattern=.pre-commit-config.yaml&q=check-executables-have-shebangs) has 1k grep.app hits. It's the final unimplemented hook used by `airflow` as mentioned in [this comment](https://github.com/j178/prek/issues/880#issuecomment-3405716137).

----

##  Notes

- **Windows**; I'm having trouble testing this on windows, since `.sh` files aren't seen as executable and setting them as executable using `git add --chmod=+x {file_name}`. See https://github.com/j178/prek/issues/928
- **Windows32**
     - Currently we do not check windows32. In this PR and the original [pre-commit implementation](https://github.com/pre-commit/pre-commit-hooks/blob/main/pre_commit_hooks/check_executables_have_shebangs.py), it takes a completely different route than UNIX based on the return value of `git config core.fileMode` (returns false only on win32`
     - 2 options:
                1. Add `win32` to the ci/cd pipeline
                2. We need to be able to overwrite the output of this command in tests (like one would in [python](https://docs.python.org/3/library/unittest.mock.html))
```rust 
# git config core.fileMode is used to determine UNIX vs win32
let status = Command::new("git")
      .arg("config")
      .arg("core.fileMode")
      .current_dir(repo_path)
      .status()?;
```

- **Bugfix** 
    -  Fixed a bug in `identify.rs`. Where the first element of a shebangs was being accessed. However the `parse_shebang` function can still return an `Ok` with an empty vector, making the access fail. 
    - This caused prek to panic when running against the [ardupilot](https://github.com/ArduPilot/ardupilot) repository.
    - I added the check in `identify.rs` but the `parse_shebang` function still contains a Todo.
    - Also described in: https://github.com/j178/prek/issues/925

### Memory changes
```bash
cargo clean
cargo build --release 
```
- Master branch:  **8.49 MB**
- Feature branch: **8.58MB**
+89,800 bytes (+1.06%) increase from master to feature branch

-----

### Performance

Ran these commands on the [ardupilot](https://github.com/ArduPilot/ardupilot) repository:
_simply because it was the top-hit on [grap.app](https://grep.app/search?f.path.pattern=.pre-commit-config.yaml&q=check-executables-have-shebangs)_

Linux
```bash
# pre-commit
pre-commit run check-executables-have-shebangs --all-files --verbose
check that executables have shebangs.....................................Passed
- hook id: check-executables-have-shebangs
- duration: 0.09s

# Old prek implementation (python)
prek run check-executables-have-shebangs --all-files --verbose
thread '<unnamed>' panicked at src/identify.rs:767:55:
index out of bounds: the len is 0 but the index is 0
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

# Old prek implementation bugfixed (python)
prek run check-executables-have-shebangs --all-files --verbose
check that executables have shebangs.....................................Passed
- hook id: check-executables-have-shebangs
- duration: 0.07s

# New implementation (rust)
rek run check-executables-have-shebangs --all-files --verbose
check that executables have shebangs.....................................Passed
- hook id: check-executables-have-shebangs
- duration: 0.01s
```
Linux output:
image

### DIFF
--- a/docs/builtin.md
+++ b/docs/builtin.md
@@ -36,6 +36,7 @@ Currently, only `https://github.com/pre-commit/pre-commit-hooks` is supported. M
 - `check-merge-conflict` (Check for merge conflicts)
 - `detect-private-key` (Detect private keys)
 - `no-commit-to-branch` (Prevent committing to protected branches)
+- `check-executables-have-shebangs` (Ensures that (non-binary) executables have a shebang)
 
 Notes:
 


feliblo:
I'm having some trouble setting `.sh` files as executable on windows, even when using git chmod. I will boot up my windows pc later and take a look.

It's almost certainly due to line 750 in `indentify.rs`:
```rust
    #[cfg(not(unix))]
    {
        executable = {
            let ext = path.extension().and_then(|ext| ext.to_str());
            ext.is_some_and(|ext| ext == "exe" || ext == "bat" || ext == "cmd")
        };
    }
```

edit: created this issue: https://github.com/j178/prek/issues/928

----

With the other bug mentioned above it could be too much for just 1 PR though.

j178:
Thank you! I'll take a look soon.

j178:
I'll try to resolve https://github.com/j178/prek/issues/928, then we can get this merge.

feliblo:
> I'll try to resolve https://github.com/j178/prek/issues/928, then we can get this merge.

I mentioned it in the other PR. But the way pre-commit resolves this in windows is, send all files to the hook and resolve for those marked as executable in the hook itself. Hence it always returning passed in the output above.

j178:
Thank you!

Issue #917, opened 2025-10-17
Binary size reporting on CI


lmmx:
Can we compute this on CI (I know CodeCov can do that automatically for JavaScript bundles or something? But not sure about Rust)

Will open an issue about that, but relative size like you did is fine 👍

_Originally posted by @lmmx in https://github.com/j178/prek/pull/915#issuecomment-3416335242_

I'll take a look at this later, but wanted to open an issue now

I built master branch as 12MB, Felix got 8 (I presume in release mode)

## Sketch

We would want a GitHub Action workflow that writes a comment on the PR comparing size diff vs trunk (master branch)

Something like

```yaml
name: Binary size check

on:
  pull_request:

jobs:
  size-diff:
    runs-on: ubuntu-latest
    steps:
      - name: Checkout PR branch
        uses: actions/checkout@v4

      - name: Install Rust
        uses: dtolnay/rust-toolchain@stable

      - name: Build PR binary
        run: cargo build --release
      - name: Get PR binary size
        id: prsize
        run: |
          SIZE=$(stat -c%s "target/release/your-crate-name")
          echo "size=$SIZE" >> $GITHUB_OUTPUT

      - name: Checkout master branch
        uses: actions/checkout@v4
        with:
          ref: master

      - name: Build master binary
        run: cargo build --release
      - name: Get master binary size
        id: mastersize
        run: |
          SIZE=$(stat -c%s "target/release/your-crate-name")
          echo "size=$SIZE" >> $GITHUB_OUTPUT

      - name: Post size comparison comment
        uses: marocchino/sticky-pull-request-comment@v2
        with:
          header: binary-size
          message: |
            📦 **Binary size comparison**
            
            - `master`: **${{ steps.mastersize.outputs.size }} bytes**
            - `PR`:     **${{ steps.prsize.outputs.size }} bytes**
            
            🔀 Difference: **${{ steps.prsize.outputs.size }} - ${{ steps.mastersize.outputs.size }} = ${{ steps.prsize.outputs.size && steps.mastersize.outputs.size && (steps.prsize.outputs.size - steps.mastersize.outputs.size) }} bytes**
```

If we can get that in Rust itself it might be simpler? Would need to look up the crates that can do that

feliblo:
Could be useful/tries to achieve something similar: https://github.com/orf/cargo-bloat-action

lmmx:
> Could be useful/tries to achieve something similar: https://github.com/orf/cargo-bloat-action

[...]

PR #927, opened 2025-10-18
perf: monitor binary size change in PRs via cargo-bloat workflow

lmmx:
- As discussed in #917 

Cargo bloat workflow to show the binary size change for PRs in the form of a "Checks" status line

> I came up with a way that shows it in the PR via a check "status" line, see [demo](https://github.com/lmmx/cargo-bloat-action-demo/pull/1) PR and [workflow source](https://github.com/lmmx/cargo-bloat-action-demo/blob/master/.github/workflows/bloat.yml)
> 
> Image 

 _Originally posted by @lmmx in [#917](https://github.com/j178/prek/issues/917#issuecomment-3418562809)_

I tried it with a status like this, (to save you having to click through to read the result) but it looks like that won't be allowed to run on forks (i.e. only j178's own PRs to this repo would have it, everyone else's would error) so I removed that and now you have to click through to the step summary to see it:

(...)

To see the result you click through to the step summary where it shows up as an "annotation" at the top and a full report below

- [x] Added mold linker and caching so (initial run took 5m)
- [x] Removed check status line (fails for PRs from other repos)
- [x] Rewrote size change calculation in Python

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -133,15 +133,14 @@ jobs:
                   raise ValueError("Could not find .text section size")
               value, unit = float(match.group(1)), match.group(2)
               multipliers = {'B': 1, 'KiB': 1024, 'MiB': 1024**2, 'GiB': 1024**3, 'TiB': 1024**4}
-              return value * multipliers.get(unit, 1)
+              size = value * multipliers.get(unit, 1)
+              return size, f"{value} {unit}"
 
           pr_text = Path('bloat_pr.txt').read_text()
           base_text = Path('bloat_base.txt').read_text()
 
-          pr_bytes = parse_size(pr_text)
-          base_bytes = parse_size(base_text)
-          pr_size = re.search(r'([\d.]+\s*[KMGT]i?B)', pr_text).group(1)
-          base_size = re.search(r'([\d.]+\s*[KMGT]i?B)', base_text).group(1)
+          pr_bytes, pr_size = parse_size(pr_text)
+          base_bytes, base_size = parse_size(base_text)
 
           pct_change = ((pr_bytes - base_bytes) / base_bytes) * 100
           pct_display = f"{pct_change:+.2f}%"


.github/workflows/bloat.yml
j178, line 1: Is it possible to move this action into `ci.yml`?

.github/workflows/bloat.yml
```diff
@@ -0,0 +1,137 @@
...
+          {
+            echo 'output<<EOF'
+            cat bloat_pr.txt
+            echo 'EOF'
+          } >> $GITHUB_OUTPUT
+
+      - name: Compare bloat results
+        id: compare
+        if: github.event_name == 'pull_request'
+        run: |
```
j178: What do you think about implementing this shell script as a Python script instead? I think that might be easier to maintain.

.github/workflows/bloat.yml
```diff
@@ -0,0 +1,137 @@
...
+            echo ""
+            echo "### PR Branch Results"
+            cat bloat_pr.txt
+            echo ""
+            echo "### Base Branch Results"
+            cat bloat_base.txt
+            echo 'EOF'
+          } >> $GITHUB_OUTPUT
+
+      - name: Display results (PR)
```
j178: I noticed the “Update commit status” step is missing from this workflow - why is that?

```
      - name: Update commit status
        if: github.event_name == 'pull_request'
        uses: actions/github-script@v7
        with:
          script: |
            await github.rest.repos.createCommitStatus({
              owner: context.repo.owner,
              repo: context.repo.repo,
              sha: context.payload.pull_request.head.sha,
              state: 'success',
              context: 'Binary Size Change',
              description: `${{ steps.compare.outputs.pct_change }} (.text: ${{ steps.compare.outputs.base_size }} → ${{ steps.compare.outputs.pr_size }})`,
              target_url: `${context.serverUrl}/${context.repo.owner}/${context.repo.repo}/actions/runs/${context.runId}`
            });
```

lmmx: I took it out as it wouldn't work for PRs from other repos, added a note in the README of the demo

> [!TIP]
> If you primarily accept PRs from forks, delete the "Update commit status" step from the workflow to avoid the skipped check.

Nice idea but suspect it wouldn't actually work for what we wanted it for :-(

.github/workflows/bloat.yml
lmmx: Yeah sure :+1:

lmmx: - [x] 3b0ee50

.github/workflows/bloat.yml
lmmx: Sure yeah, a Python script in `.github/scripts/` would work

j178: Thanks!

lmmx: In fact you can put a Python program directly in here, set `shell: Python`. Would this be OK?

- [x] f5027bb

lmmx: Marked as resolved but left a comment below

lmmx:
Rewrote the bloat check in Python as `shell: python` (i.e. in the CI workflow), let me know if you'd rather have it as a script

The check status line worked in my demo (because it was a PR from the same repo) but unfortunately failed when I tried it here (as it was coming from repo `lmmx/prek` to `j178/prek`, GitHub's security model only allows same-repo PRs to use that API)

.github/workflows/bloat.yml
j178: > In fact you can put a Python program directly in here, set `shell: Python`. Would this be OK?
> 
> 
> 
> - [x] f5027bb

Yeah, that’s exactly what I meant too.

j178:
This looks great!

image

lmmx:
Thanks for the edits!

lmmx:
I’ve gone and rebased the branches I had open and added their outputs (the annotation section at the bottom of the Summary page with the 🏠 icon)

IMG_4602

and pasted them into the PR description like this

> [Binary Size Change](https://github.com/j178/prek/actions/runs/18629114297/job/53111274368#step:9:312)
> +0.63% (.text: 16.0 MiB → 16.1 MiB)

PR #942, opened 2025-10-20
Workaround PR from forks can't comment

j178:


potiuk:
Just a word of warning @j178 -> this is generally a dangerous path. We completely banned all kinds of workflow priviledge escallation (workflow-run from PRs and especially ESPECIALLY pull-request-target kinds of workflow) because they are inherenty insecure and there are **multiple** ways how potentially that might give users creating PRs from forks potentially gaining access to secrets, caches. This can happen in many ways and there were a number of security issues triggered by it.

I have not seen anything suspicious in this workflow, but I've learned over time that there are many unobvious ways how malicious users could exploit it.

Example here: https://www.stepsecurity.io/blog/github-actions-pwn-request-vulnerability 

Also we had our own vulnerability that I wrote about here: https://medium.com/apache-airflow/unraveling-the-code-navigating-a-ci-release-security-vulnerability-in-apache-airflow-620214a96297 (unavailable currently due to global AWS outage now).

At the very least I strongly recommend applying CodeQL and https://github.com/zizmorcore/zizmor to analyse the workflow security.

j178:
Thanks for the reminder @potiuk! I borrowed this straight from ruff, and since they've got way more influence than prek, I think this might be okay for now? I'll definitely add the zizmor workflow later though.

potiuk:
> Thanks for the reminder @potiuk! I borrowed this straight from ruff, and since they've got way more influence than prek, I think this might be okay for now? I'll definitely add the zizmor workflow later though.

Well. I just run zizmor on the current actions now and I think you want to do it sooner rather than later:


```
zizmor --gh-token=$(gh auth token) j178/prek
```

image

potiuk:
This is what I mean by "insecure workflow":

image

Issue #928, opened 2025-10-18
[bug] Executables not correctly identified on windows


Felix-Blom:
### Summary

Mentioned here: https://github.com/j178/prek/pull/924#issuecomment-3418647586  on https://github.com/j178/prek/pull/924

Executables on windows are currently recognized by:
```
    #[cfg(not(unix))]
    {
        executable = {
            let ext = path.extension().and_then(|ext| ext.to_str());
            ext.is_some_and(|ext| ext == "exe" || ext == "bat" || ext == "cmd")
        };
    }
```

However its also possible to set files as executable using `git add chmod +x` as explained [here](https://stackoverflow.com/questions/21691202/how-to-create-file-execute-mode-permissions-in-git-on-windows).

I will still need to compare but the [pre-commit implementation](https://github.com/pre-commit/pre-commit-hooks/blob/main/pre_commit_hooks/check_executables_have_shebangs.py) of above-mentioned PR expects .sh files are being checked when marked as executable in git.

### Platform

Windows (any)

### Version

prek 0.2.9

j178:
pre-commit uses `os.access(path, os.X_OK)` to check if a file is executable, but here's the thing - on Windows, this actually returns true for any file that's accessible.

https://github.com/pre-commit/identify/blob/a07859b08789d088e4b8804a6185e16f57e4763f/identify/identify.py#L56-L60

Under the hood, `os.access` gets translated to this code on Windows:

https://github.com/python/cpython/blob/2699643d715aeccdbdbb9b429f7bd423050e8f41/Modules/posixmodule.c#L3329-L3346

feliblo:
@j178 You're right, it seems quite confusing on `pre-commit` aswell. I think their implementation is actually kind of broken.

**Long text, TLDR:** 
1. pre-commit acts weird on windows (based on `git config core.fileMode`)
2. prek only needs to add the `git ls-files --stage` files marked as executable to the `executable` variable (on windows). 
3. prek needs to make sure windows executables that can't contain shebangs (e.g. `.bat`) are excluded from individual hooks e.g. `check-executables-has-shebangs` hook (by simply utilizing/documenting the - exlude key for that hook).

 This way:

  - You get a complete list of actual executables (for the possibility of other hooks)
  - You create a sublist of executables that can contain shebangs for shebang related hooks.


I can pick this up. Do you agree?

## pre-commit behaviour

None of these files actually contain shebangs. When not tagging any file as executable using git:
```sh
# Windows core fileMode false
[...]

PR #945, opened 2025-10-20
Add git tagged executables when on Windows

feliblo:
Closes: https://github.com/j178/prek/issues/928

Adds all files in windows that have been commited as executable using either `git update-index --chmod=-x path/to/file` or `git add --chmod=+x path/to/file`. 

@j178 Not sure whether the test positioning is correct

### DIFF
--- a/src/identify.rs
+++ b/src/identify.rs
@@ -749,10 +749,34 @@ pub(crate) fn tags_from_path(path: &Path) -> Result<Vec<&str>> {
     }
     #[cfg(not(unix))]
     {
-        executable = {
-            let ext = path.extension().and_then(|ext| ext.to_str());
-            ext.is_some_and(|ext| ext == "exe" || ext == "bat" || ext == "cmd")
-        };
+        let output = std::process::Command::new("git")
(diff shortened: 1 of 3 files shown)


src/identify.rs
```diff
@@ -749,10 +749,34 @@ pub(crate) fn tags_from_path(path: &Path) -> Result<Vec<&str>> {
     }
     #[cfg(not(unix))]
     {
-        executable = {
-            let ext = path.extension().and_then(|ext| ext.to_str());
-            ext.is_some_and(|ext| ext == "exe" || ext == "bat" || ext == "cmd")
-        };
+        let output = std::process::Command::new("git")
```
j178, line 752: This makes performance unacceptable - every file now takes 100+ms just to get its tags. I'd say we should just stick with the pre-commit behavior for now.

feliblo: I think this perfomance is the reason that this hook always runs in pre-commit. So you can simply check the git committed executables within the hook, instead of skipping the hook entirely. It still isn't behaving similarly in prek.

PR #947, opened 2025-10-21
Hardening github actions

j178:
- Pin action to commits
- Add zizmor action
- Set `persist-credentials: false` for `actions/checkout`

### DIFF
--- a/.github/workflows/bloat-comment.yml
+++ b/.github/workflows/bloat-comment.yml
@@ -54,6 +54,8 @@ jobs:
       - name: Comment on PR with bloat results
         uses: actions/github-script@ed597411d8f924073f98dfc5c65a23a2325f34cd # v8.0.0
         if: steps.download-bloat-check.outputs.found_artifact == 'true'
+        env:
+          PR_NUMBER: ${{ steps.pr-number.outputs.pr-number }}
         with:
           script: |
             const fs = require('node:fs');
@@ -62,12 +64,12 @@ jobs:
             const body = `${marker}\n${comparison}`;
 
             const { repo, owner } = context.repo;
-            const issue_number = context.issue.number;
+            const prNumber = Number(process.env.PR_NUMBER);
 
             const comments = await github.paginate(github.rest.issues.listComments, {
               owner,
               repo,
-              issue_number,
+              issue_number: prNumber,
               per_page: 100,
             });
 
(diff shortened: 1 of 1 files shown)


potiuk:
One other small comment. We use "octopin" from the eclipse foundation (and with monorepo support of prek it is nicely separate in a separate .pre-commit.yaml - see https://github.com/apache/airflow/blob/main/dev/.pre-commit-config.yaml - to nicely keep hash-commits updaed for our actions with comments about the version used. Also recommended.

lmmx:
Great work here @j178 :clap:

You can also enforce SHA pinning as repo policy now (you may already have)

> ### Enforce SHA pinning
> 
> To proactively limit the impact of a compromised dependency, GitHub recommends that workflows [pin dependency versions to a specific commit SHA](https://docs.github.com/actions/reference/security/secure-use#using-third-party-actions). This will prevent malicious code added to a new or updated branch or tag from being automatically used. Administrators can now enforce the use of SHA pinning through the allowed actions policy. A new checkbox appears under each radio selection, except when actions are disabled at the enterprise, organization, or repository level. The policy will check for a full commit SHA, and any workflow that attempts to use an action that isn’t pinned will fail.

_via [GitHub Actions policy now supports blocking and SHA pinning actions](https://github.blog/changelog/2025-08-15-github-actions-policy-now-supports-blocking-and-sha-pinning-actions/#enforce-sha-pinning) (August 15, 2025)_

.github/zizmor.yml
```diff
@@ -0,0 +1,14 @@
+# Configuration for the zizmor static analysis tool, run via pre-commit in CI
```
ulgens, line 1: This seems to be misleading. zizmor is run via its own action and currently there is no clear path to run it locally.

.github/workflows/zizmor.yml
ulgens, line 1: Coming from https://github.com/j178/prek-action/pull/55, was there a reason to prefer a new workflow over the git hook?

j178: zizmor action supports GitHub advanced security and add inline annotations to changes, so I would prefer using zizmor action.

.github/zizmor.yml
j178: Yeah, this comment should be updated.

.github/workflows/zizmor.yml
ulgens: Got it. With this setup, what should one do to run it locally?

j178: I'll just let CI checks it :)

ulgens: Fair enough 😄

PR #949, opened 2025-10-21
Support `language_version: system`

### LINKED ISSUE #698: "Invalid `language_version` value: `system`" encountered with `npm`-based local hooks

j178:
Closes #698 
Supersedes #898 

Correctly supports `language_version: system`, which uses whatever toolchain is already installed on the system, without trying to download a new one.

### DIFF
--- a/src/languages/python/python.rs
+++ b/src/languages/python/python.rs
@@ -249,6 +249,7 @@ impl Python {
                             "No suitable system Python version found and downloads are disabled"
                         );
                     }
+
                     debug!(
                         "Retrying venv creation with managed Python downloads: `{}`",
                         info.env_path.display()
--- a/src/languages/version.rs
+++ b/src/languages/version.rs
@@ -41,7 +41,7 @@ impl LanguageRequest {
     pub(crate) fn allows_download(&self) -> bool {
         match self {
             LanguageRequest::Any { system_only } => !system_only,
-            _ => false,
+            _ => true,
         }
     }
 


lmmx:
Ah thanks for stepping in with this 😄

Issue #955, opened 2025-10-23
[Perf] CI speed profiling with hotpath


lmmx:
Since we have the cargo bloat (#917), maybe we could do performance profiling in the same way with [hotpath](https://github.com/pawurb/hotpath) (which I just found out about!)

It profiles both latency and allocation, maybe this could be useful to pursue the perf goals here?

I'd set it up with a demo suite to be used to benchmark and cache the build as done for the cargo-bloat job.

The open question here is which pre-commit hooks to use as a 'standard' benchmark measure on...

1. If you use a small number of built-in hooks then run-time will be dominated by setup
2. If you use a larger number of external hooks then it'll be from running those

I'd pick 1 as it'd measure things directly in control of, and would likely be more sensitive (show small slowdowns as a bigger +% change overall), but open to others' feedback

## Example

An example of it coming in handy from [the author’s Twitter](https://x.com/_pawurb/status/1979558739998036290):

<img width="746" height="222" alt="Image" src="https://github.com/user-attachments/assets/4d489cd2-9875-4509-93c9-bbefc857e88f" />

<img width="877" height="105" alt="Image" src="https://github.com/user-attachments/assets/2cbf4133-16f3-41c0-9505-db93040fd241" />

lmmx:
> > > <img width="1524" height="858" alt="Image" src="https://github.com/user-attachments/assets/52e4895d-0892-40fc-8e15-ae0477b48714" />

N/A in the value means it's failing to get a value at all

Also see this in the CI summary itself

> > > <img width="630" height="531" alt="Image" src="https://github.com/user-attachments/assets/20a70e6a-9c18-429e-b36f-392d62fa558a" />

Detailed output blocks both blank

> > > <img width="602" height="435" alt="Image" src="https://github.com/user-attachments/assets/14fcb2ec-a868-466f-a6f9-a93b6dd2b809" />

PR #956, opened 2025-10-23
Profile latency and allocations with `hotpath`

lmmx:
> - #955 

- **perf(hotpath): profile prek run on its own repo**

## Hotpath instrumentation

I've added `#[cfg_attr(feature = "hotpath", hotpath::measure)]` to the main hot paths I think we care about:

- `run()` - the entry point for `prek run`
- `collect_files()` - file discovery and filtering
- `install_hooks()` - hook environment setup
- `run_hooks()` - overall hook execution
- `run_hook()` - individual hook runs
- `clone_repo()` - repo cloning

The guard is initialised in `main()` so profiling output prints when the program exits.

**Status**: ready for review

- The CI is failing because the PR is not merged and the new check will only run from the master branch

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -196,7 +196,7 @@ jobs:
 
       - name: Display results (PR)
         if: github.event_name == 'pull_request'
-        run: cat comparison.txt >> $GITHUB_STEP_SUMMARY
+        run: cat bloat_comparison.txt >> $GITHUB_STEP_SUMMARY
 
       - name: Display results (Push)
         if: github.event_name == 'push'


lmmx:
Hmmm I'm getting an error here

```
Run # Build and profile base branch
Previous HEAD position was ca9423d Merge b3cee9ffa89779b707b8db6f7842a8cd4c180642 into 697b101b6d20a3d4b2d2dc6258d486c8402a842c
HEAD is now at 697b101 Improve styling for stashing error message (#953)
error: the package 'prek' does not contain this feature: hotpath
Error: Process completed with exit code 101.
```

but I added that feature to Cargo TOML

- It is on 697b101 which is the master branch HEAD commit without the feature, hence the error (this just means it won’t run a modified workflow introduced by a commit

In other words it is failing because it is not merged (so would need to be merged to see it in action)!

When it succeeds (blocked by this step) it should give a comment like the cargo-bloat one (uses the same mechanism)

j178:
Thanks! But I'll put this on the back burner for now and come back to it after I wrap up some other higher priority stuff.

lmmx:
Yes no urgency ☺️ I just saw it cross my timeline and thought it could be good for perf chasing, but already done lots on this lately

j178:
Thank you!

lmmx:
I don't think it's working correctly, will have a look into it later

> - (let's track in #955) since that is still open

PR #957, opened 2025-10-23
Allow reqwest to use system tls

st1971:
Trying to replace pre-commit with prek in a work environment where custom self signed certificates are required due to tls inspection, this PR adds an additional feature (native-tls) to the reqwest crate and additional environment variable PREK_REQWEST_SYSTEM_TLS, when this variable is set to true reqwest will use system trust store rather than the one packaged in rustls-tls-webpki-roots.  This allows the auto installation of UV, Node etc in an environment where custom trusted certificates are required for download.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -484,6 +484,22 @@ dependencies = [
 ]
 
 [[package]]
+name = "core-foundation"
+version = "0.9.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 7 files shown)


j178:
Hi, could I ask why this PR was closed?

st1971:
Hi, Sorry after i raise the PR realised the change was wrong so closed this, will be reworking this today and will re-raise when its working correctly.

PR #959, opened 2025-10-24
Allow using system trusted store by `PREK_NATIVE_TLS`

st1971:
update reqwest cargo import to add feature rustls-tls-native-roots, this allow prek to work in an enviroment where additional system level trusted certificates are required, feature rustls-tls-native-roots respected the SSL_CERT_FILE env var.

Verified on macos / linux only.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -45,7 +45,16 @@ Prek supports the following environment variables:
 - `PREK_ALLOW_NO_CONFIG` — Allow running without a .pre-commit-config.yaml (useful for ad‑hoc runs).
 - `PREK_NO_CONCURRENCY` — Disable parallelism for installs and runs.
 - `PREK_NO_FAST_PATH` — Disable Rust-native built-in hooks; always use the original hook implementation. See [Built-in Fast Hooks](builtin.md) for details.
-- `PREK_UV_SOURCE` — Control how uv (Python package installer) is installed. Options: `github` (download from GitHub releases), `pypi` (install from PyPI), `tuna` (use Tsinghua University mirror), `aliyun` (use Alibaba Cloud mirror), `tencent` (use Tencent Cloud mirror), `pip` (install via pip), or a custom PyPI mirror URL. If not set, prek automatically selects the best available source.
+- `PREK_UV_SOURCE` — Control how uv (Python package installer) is installed. Options:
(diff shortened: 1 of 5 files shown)


j178:
I noticed that in `uv`, you have to pass the `--native-tls` flag to enable native TLS roots, because "reading the native certificates can add hundreds of milliseconds to client initialization." For performance reasons, should we maybe make this opt-in too? Check out https://github.com/astral-sh/uv/pull/2362

st1971:
the PR that was highlighted is no longer used in UV, however, looked at the UV code and have used a similer method to what is current in use.  If SSL_CERT_FILE is set to a file that exists or PREK_NATIVE_TLS is set to true then system tls trust store will be used, if these are not set it will fall back to webpki-roots.

the change has dropped the test coverage level, however to put some meaninful tests on this change we would need to update the environment table during testing which would require making the test single threaded.

j178:
> put some meaninful tests on this change we would need to update the environment table during testing which would require making the test single threaded

Since we call prek in a subprocess during integration tests, I assume it's safe to use subprocess-specific environment variables here?

Could you also document the new environment variables in `docs/configuration.md`?

st1971:
Sorry, struggling a little with this, to be honest this change should be covered using unit test which as we should really be testing that the returned client can access an https end point when configured in different ways, which is only possible if the unit tests are single threaded as the mutation of the environment table is not safe, could do some stuff with extra locks if the EnvVars type was in the same scope as the main binary code.

Testing this within integration tests, not really sure how to make this work, within the integration test cannot construct a reqwest client instance so would have to set-up multiple language tests just to trigger downloads, which seems somewhat excessive.

What i could do is move the reqwest client creation and resource downloading code into a lib, in the same way EnvVars is, in this way it can be tested independently of the rest of the code.

Thoughts?

j178:
I extracted the env var reading from `get_reqwest_client`, so now we can create a native TLS client without messing with the process env table. I also added a basic test just to make sure the native TLS client actually works.

And I took the opportunity to do a bit more refactoring by making the client a global shared instance - hope that makes sense.

st1971:
Nice, had been messing about moving the client / download_and_extract to a lib which also works well from a test perspective, but yours is cleaner.  if your happy with it then i am.

its really hard to get reqwest to error even when you want it to!

PR #968, opened 2025-10-26
support Deno auto installation

janosh:
`deno` language support for pre-commit hooks with automatic version management from GitHub releases.

**Features:**
- Auto-downloads requested Deno versions or uses system installation (if version matches)
- Supports `language_version` (exact, semver ranges) and `additional_dependencies` (npm:, jsr:)
- Creates isolated `DENO_DIR` for dependency caching
- Cross-platform support (Linux, macOS, Windows)

**Changes:**
- Add `DenoInstaller`, `DenoVersion`, and `DenoRequest` for version management
- Add Deno to storage buckets and language registry
- Add 8 comprehensive tests
- Update documentation

Mirrors Node.js/Go installation patterns for consistency.

### DIFF
--- a/src/languages/deno/deno.rs
+++ b/src/languages/deno/deno.rs
@@ -19,12 +19,9 @@ use crate::store::{CacheBucket, Store, ToolBucket};
 /// Find the script in the entry that should be cached.
 /// Handles both direct scripts and `deno run ...` commands.
 fn find_script_to_cache(entry: &[String]) -> Option<&str> {
-    if entry.is_empty() {
-        return None;
-    }
+    let first = entry.first()?.as_str();
 
-    // Check if first arg is a built-in Deno command
-    let first = entry[0].as_str();
+    // Skip built-in Deno commands
     if matches!(
         first,
         "fmt" | "lint" | "test" | "check" | "bundle" | "doc" | "repl" | "eval"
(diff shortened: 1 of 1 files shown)


j178:
wow!

src/languages/deno/deno.rs
```diff
@@ -53,44 +27,74 @@ impl LanguageImpl for Deno {
...
+            // Check if repo has deno.json to copy, otherwise create minimal one
+            let mut needs_deno_json = true;
+            if let Some(repo_path) = hook.repo_path() {
+                let repo_deno_json = repo_path.join("deno.json");
+                if repo_deno_json.exists() {
+                    // Copy the deno.json from the repo
+                    fs_err::tokio::copy(repo_deno_json, &deno_json).await?;
+                    needs_deno_json = false;
+                }
+                // Deno can run scripts directly from the repo without installation
```
j178: I think we need to cache dependencies from `deno.json` (by `deno install`?). The idea is that `install` should set up everything needed for the hooks to run, so when we call `run`, it can work completely offline.

tests/languages/deno.rs
```diff
@@ -0,0 +1,365 @@
...
+    - duration: [TIME]
+      deno X.X.X
+      v8 X.X.X.X
+      typescript X.X.X
+
+    ----- stderr -----
+    "#);
+}
+
+/// Test that when deno is not available, a helpful error is shown.
```
j178: `deno` will be installed automatically, right?

src/languages/deno/deno.rs
```diff
@@ -129,17 +134,24 @@ impl LanguageImpl for Deno {
...
-        let entry = hook.entry.resolve(None)?;
+        let entry = hook.entry.split()?;
         let run = async move |batch: &[&Path]| {
-            let mut output = Cmd::new(&entry[0], "deno run")
+            // Replace "deno" with the actual path to the installed deno binary
+            let command = if entry[0] == "deno" {
+                deno_bin.as_path()
+            } else {
+                Path::new(&entry[0])
+            };
```
j178: How about symlinking the deno executable into the hook env bin dir and prepending that bin dir to PATH, like we did in the `node` implementation? That might work better than this approach. For example, it would work for cases like `entry: ./run_deno.sh`.

j178: If you agree, I can take care of it.

src/languages/deno/deno.rs
janosh: not an expert of `deno`, just a user. but i think `deno` maintains it's own global cache of every package that was ever imported in an executed script (be it from NPM, JSR or directly via URL from GitHub)

janosh: i guess we can call `deno cache` during hook installation. subtle benefit of that:
- with `deno cache`: Dependencies are downloaded during `prek install` (slower install, faster first run, can detect dependency issues early)
- without `deno cache`: Dependencies are downloaded on first hook run (faster install, slightly slower first run)

janosh:
btw, feel free to make changes as you see fit in this PR. i'm outside my comfort area here and may have made stupid mistakes so everything is up for scrutiny. also, feel free to merge this PR into the `deno` branch and continue work there

j178:
I'm going to merge this into `deno` branch and continue there, thank you!

PR #980, opened 2025-10-27
Treat every file as executable on Windows to keep compatibility with pre-commit

### LINKED ISSUE #928: [bug] Executables not correctly identified on windows

j178:
Closes #928

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -23,7 +23,7 @@ repos:
       - id: cargo-fmt
         name: cargo fmt
         entry: cargo fmt --
-        language: perl
+        language: system
         types: [rust]
         pass_filenames: false # This makes it a lot faster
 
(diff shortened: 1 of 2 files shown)


feliblo:
@j178 after this is closed I can fix the shebang hook!

PR #991, opened 2025-10-29
Remove hotpath profiling

### LINKED ISSUE #955: [Perf] CI speed profiling with hotpath

j178:
The profiling isn’t very useful right now - it’s way too sensitive to noise. What we really need is proper benchmarking that runs multiple times to filter out all that noise, like what you get with https://codspeed.io/ or https://bencher.dev/

Closes #955

### DIFF
--- a/.github/workflows/performance.yml
+++ b/.github/workflows/performance.yml
@@ -9,14 +9,6 @@ concurrency:
 
 permissions: {}
 
-env:
-  UV_VERSION: "0.8.22"
-  NODE_VERSION: "20"
-  GO_VERSION: "1.24"
-  PYTHON_VERSION: "3.12"
-  LUA_VERSION: "5.4"
-  LUAROCKS_VERSION: "3.12.2"
-
 jobs:
   bloat-check:
     runs-on: ubuntu-latest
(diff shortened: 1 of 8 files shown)


j178:
@lmmx What do you think?

lmmx:
Easy come, easy go! Fine by me, experiments are good (:

lmmx:
It is zero compile cost so could just turn off the CI bot and document it exists for future perf chasers.

j178:
I want to remove it because that cleaned up the dependency tree a lot :)

https://github.com/j178/prek/pull/991/files#diff-13ee4b2252c9e516a0547f2891aa2105c3ca71c6d7a1e682c69be97998dfc87e

Issue #43, opened 2024-11-11
Support `ruby` language


j178:

is-alnilam:
I have a draft of code to support Ruby, but would like to check a few things to finish it. Basically it comes down to versions. There are a number of Ruby version managers out there on Linux, such as `rvm`, `rbenv`, `asdf`, `mise`, `chruby`. Then there's `homebrew` (particularly on Mac), or `rv` coming along (and already able to do the basics of installing many recent versions). On Windows, the situation's a mess, needing manual installation.

What I've currently implemented ignores adding new versions, and instead looks in `$PATH` (potentially finding more than one Ruby version), then also hunts through standard locations. On Windows it could either rummage through the registry, or use the `ridk` tool—but at the moment it's `%PATH%` only (again, capable of handling multiple versions). If no suitable Ruby is found, it just prompts the user, suggesting a suitable command if they have a manager installed.

Does that sound of use? Or should I cut back the hunting and prompting, and/or is it necessary to implement some form of downloader?

My team need to be able to run the `rubocop` hook on Windows, so I can hopefully get them to test 'real' use, although running the test suite may be harder. I've been successfully running this on Linux. The tests will need a bit more work before a PR, and I'll want to work out the coverage and bloat details too.

j178:
It's fine to leave out the downloader - we can always add it later.

MRigal:
`rv` although very new, as a "copy" of `uv`, may be the way to go for the downloader, but indeed that would be limited to Linux and Mac at the beginning, so probably good to be able to make use of what is available on `$PATH` to start!

j178:
I might just want to “reuse” (okay, steal) some of the downloader code from rv instead of relying on its binary. Ref #765

is-alnilam:
I had a quick try at that approach, and was finding it awkward to parse the available versions. But yes, that does seem a good option!

is-alnilam:
Pull request #993 is a partial fix for Ruby support (as mentioned, there's no support for downloading new Ruby builds just yet). Fixing the CI (particularly on Windows) and getting the test coverage up took a bit longer than I'd hoped...

There's a fair amount of code to hunt for Rubies installed by RVM, rbenv, mise, asdf, homebrew, etc. That could be trimmed out if preferred, but it may help to keep the utility high until downloading can be added.

MRigal:
Should this issue be closed now that https://github.com/j178/prek/releases/tag/v0.2.13 added the partial support for Ruby?

[...]

PR #993, opened 2025-10-29
Add Ruby support (no download support yet)

is-alnilam:
Currently only pre-existing Rubies are used, although the code will search for and consider multiple versions. This allows a hook to require 3.4.7 when the default system Ruby is 3.4.6, but RVM (or a similar tool) has installed 3.4.7. If a suitable Ruby cannot be found, the user is prompted to install one manually and re-run.

Hooks get an isolated GEM_HOME, based on the hook repository, any additional_dependencies, and the interpreter version/path. Any Ruby upgrade would therefore re-install the gems to ensure that stale versions aren't used. Repository gem dependencies are installed based on the included gemspec files, combined with the additional_dependencies field.

Updates #43

### DIFF
--- a/.github/workflows/build-binaries.yml
+++ b/.github/workflows/build-binaries.yml
@@ -45,7 +45,7 @@ jobs:
           command: sdist
           args: --out dist
       - name: "Upload sdist"
-        uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4.6.2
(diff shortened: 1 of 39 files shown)


is-alnilam:
Just to say that I'll have limited time after next week to look at this for a while. All your tweaks look good to me (not that you need my approval!) and thanks for finding the areas where I'd missed parts of the existing code...

Let me know if you want me to pick up on any tasks over the next few days, for instance rebasing / catchup merge from master, otherwise I'll leave it in your hands so we don't duplicate effort.

j178:
Thank you!

PR #1018, opened 2025-11-02
Support rootless container

st1971:
attempts to autodetect if the container runtime is rootless, if it
cannot be detected then will assume rootful for docker and rootless for
podman as these are the most common way to install.

in addition environment variable PREK_CONTAINER_RUNTIME_ROOTLESS has
also been provided to allow detection to be overridden if required, this
variable should be treated as a boolean.

### DIFF
--- a/crates/prek-consts/src/env_vars.rs
+++ b/crates/prek-consts/src/env_vars.rs
@@ -22,7 +22,6 @@ impl EnvVars {
     pub const PREK_NATIVE_TLS: &'static str = "PREK_NATIVE_TLS";
     pub const SSL_CERT_FILE: &'static str = "SSL_CERT_FILE";
     pub const PREK_CONTAINER_RUNTIME: &'static str = "PREK_CONTAINER_RUNTIME";
-    pub const PREK_CONTAINER_RUNTIME_ROOTLESS: &'static str = "PREK_CONTAINER_RUNTIME_ROOTLESS";
 
     // PREK internal environment variables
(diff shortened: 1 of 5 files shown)


j178:
Thanks! Would you mind splitting this into smaller PRs, with each one solving a single issue? It'd be much easier to review and test.

st1971:
Sure no problem, may take me a few days, busy work week this week :-)

Do you want me to close this one?

j178:
> Do you want me to close this one?

We can keep this open and focus on rootless container support here since the branch name is “feat/rootless-containers”

st1971:
I'm looking at this again and given the potential issue auto detecting the state of the current runtime, I am starting to think it may be better to simply provide an environment variable `PREK_CONTAINER_RUNTIME_ROOTLESS` as a boolean to drive this.  The user should know if they are running rootless or not, we could make some assumptions based on the runtime i.e podman defaults to rootless as this is default for a podman install and docker defaults to root base as this is the default for a docker install.  would be good to get your thoughts?

j178:
Rootless mode is something we can detect with 100% accuracy, right? If so, maybe it doesn't need to be a user option.

st1971:
have just re-based and pushed changed to this, it will attempt to auto detect, if this fails will default to rootless for podman and rootful for docker (these are the common install patterns) have also included an override env var (optional) to force rootless / rootful if needed.  hope this is ok.

PR #1030, opened 2025-11-03
refactor parse boolish

st1971:
refactor parse boolish, moved into prek-const, add additional convenience method to read env var and treat as bool


j178:
Thanks!

PR #1033, opened 2025-11-03
Support `PREK_CONTAINER_RUNTIME=podman` to override container runtime

st1971:
use both docker and podman runtimes

1. if only docker on path use docker runtime
2. if only podman on path use podman runtime
3. if both docker and podman on path prefer docker
4. If PREK_CONTAINER_RUNTIME set use that runtime, currently supports docker and podman

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -29,9 +29,9 @@ Specify the minimum required version of prek for the configuration. If the insta
 
 Example:
 
-```yaml
-minimum_prek_version: "0.2.0"
-```
+  ```yaml
+  minimum_prek_version: '0.2.0'
+  ```
 
 The original `minimum_pre_commit_version` option has no effect and gets ignored in prek.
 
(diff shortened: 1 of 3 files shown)


docs/configuration.md
```diff
@@ -47,19 +47,20 @@ Prek supports the following environment variables:
...
-    - `tencent` (use Tencent Cloud mirror)
-    - `pip` (install via pip)
-    - a custom PyPI mirror URL
-
-    If not set, prek automatically selects the best available source.
+  If not set, prek automatically selects the best available source.
 
 - `PREK_NATIVE_TLS` - Use system's trusted store instead of the bundled `webpki-roots` crate.
 
+- `PREK_FORCE_PODMAN` - If both docker and podman exist force the use of podman over docker.
```
j178: I’m thinking we could call this `PREK_CONTAINER_RUNTIME`. By default, prek would automatically try whatever’s available—docker, podman, etc. But you could override that with `PREK_CONTAINER_RUNTIME` if you want to force a specific one. Like, set it to `docker` to only use docker, or `podman` to only use podman. What do you think?

st1971: That makes sense to me, happy to make that change later today 👍

st1971: Have updated as suggested, hope it looks ok :-)

j178:
Thank you!

st1971:
No problem, should be thanking you for the goodness knows how many hours you have spent on this :-)

PR #1058, opened 2025-11-10
Use temp dir for ruby test

yihong0618:
same as https://github.com/yihong0618/prek/commit/c1e69ce1af393ea49dd44c7f896637a21f2a4105


j178:
Thanks!

PR #1073, opened 2025-11-14
Support `language: unsupported` and `language: unsupported_script` introduced in pre-commit v4.4

### LINKED ISSUE #1070: language: unsupported and unsupported_script are not supported
### Summary

The [pre-commit documentation](https://pre-commit.com/#unsupported) says `language: system` has been replaced by `language: unsupported` and that `system` will be removed in a future version.

Similarly, `language: script` is now `language: unsupported_script`.

However, prek does not seem to recognize either `unsupported` or `unsupported_script`.

Thanks!

### Platform

Darwin 24.5.0 arm64

### Version

prek 0.2.13 (9c70b5093 2025-11-04)

### .pre-commit-config.yaml

```yaml
minimum_prek_version: 0.2.13
repos:
  - repo: local
    hooks:
    - id: format
      name: Format code
      entry: make format
      language: unsupported
      pass_filenames: false
```

### Log file

```
2025-11-13T15:19:20.689029Z DEBUG prek: 0.2.13 (9c70b5093 2025-11-04)
2025-11-13T15:19:20.689127Z DEBUG Args: ["prek", "run"]
2025-11-13T15:19:20.715383Z DEBUG Git root: /Users/ryotaushio/python
2025-11-13T15:19:20.715565Z TRACE Executing `/usr/local/bin/git ls-files --unmerged`
2025-11-13T15:19:20.738260Z DEBUG Found workspace root at `/Users/ryotaushio/python`
2025-11-13T15:19:20.738293Z TRACE Include selectors: ``
2025-11-13T15:19:20.738300Z TRACE Skip selectors: ``
2025-11-13T15:19:20.738375Z DEBUG discover{root="/Users/ryotaushio/python" config=None refresh=false}: Performing fresh workspace discovery
2025-11-13T15:19:20.748975Z DEBUG Loading project configuration path=.pre-commit-config.yaml
2025-11-13T15:19:20.754374Z TRACE discover{root="/Users/ryotaushio/python" config=None refresh=false}: close time.busy=16.0ms time.idle=5.46µs
```

j178:
Honestly, I have no idea why `system` and `script` got deprecated or what we're supposed to use instead. I probably won't deprecate them in prek.

Closes #1070

### DIFF
--- a/tests/languages/unsupported.rs
+++ b/tests/languages/unsupported.rs
@@ -1,10 +1,10 @@
-use crate::common::{TestContext, cmd_snapshot};
-use assert_fs::fixture::{FileWriteStr, PathChild};
-
 /// Test `language: unsupported` and `language: unsupported_script` works.
 #[cfg(unix)]
 #[test]
 fn unsupported_language() -> anyhow::Result<()> {
+    use crate::common::{TestContext, cmd_snapshot};
+    use assert_fs::fixture::{FileWriteStr, PathChild};
+
     let context = TestContext::new();
     context.init_project();
     context.write_pre_commit_config(indoc::indoc! {r"


nathanjmcdougall:
@j178 

> I have no idea why system and script got deprecated or what we're supposed to use instead.

I think the motivation is basically that the name is misleading - `language: system` implies there's a language supported by pre-commit called `system`, whereas really this is an escape hatch, and there's no language management features at all. Likewise, for `language: script`, it could be misinterpreted. I'm guessing the folks working on pre-commit were getting a lot of support requests from people misunderstanding the degree to which `language: system` should "just work" out of the box.

To be clear, the deprecation only applies to the name used for this config. There's no plan to deprecate the new `language: unsupported` and `language: unsupported_script` specifiers.

The irony here is that the new names are probably misleading too! They imply that those two options are _themselves_ going to be deprecated at some point in the future.

Regardless of the motivations, I would personally support that in `prek` you avoid breaking backward compatibility over a naming issue.

j178:
> To be clear, the deprecation only applies to the name used for this config. There's no plan to deprecate the new language: unsupported and language: unsupported_script specifiers.

Thanks, that makes more sense now.

PR #1116, opened 2025-11-19
docs: refer airflow as Apache Airflow

Lee-W:
According to the [Apache Product Name Usage Guide](https://www.apache.org/foundation/marks/guide#naming), we would like to refer to `Airflow` as `Apache Airflow`. It would be appreciated if we could get this updated. Thanks!


j178:
Thanks!

PR #1129, opened 2025-11-21
Support orphan projects

### LINKED ISSUE #1137: Inability for workspace mode to only process files one time when nested in subprojects
In your Workspace docs, you write:

> Note: Files in subprojects will be processed multiple times - once for each project in the hierarchy that contains them. For example, a file in src/backend/ will be checked by hooks in src/backend/, then src/, then the workspace root.

Is there a way to isolate subprojects automatically when running prek? I know this can be done via --skip flags manually, but if I'm working on a project with many subprojects that I don't want hooks to run twice for, this can be annoying to implement.

For example, when running prek with ruff hooks in both the root level and in subprojects with different ruff hook versions, if deeper projects have code that conflicts with the root level ruff hook, even though the ruff hook passes for the subproject, it doesn't pass when files are checked again with a different version of the ruff hook at the root level.

I guess I could either remove certain hooks from the subproject configurations or from the root level. But keeping them at the subproject level is appealing for workflow: when I'm working on a certain subproject, it makes more sense for me to update hooks while I'm there and fix any issues that arise. But I wouldn't want to always have to update hooks and any arising issues for the entire repo.

And removing hooks from the root level doesn't make sense, because there are some files that aren't in other packages that I'd like to run prek on.

Is there a reason behind this design choice I'm missing? Or do you think adding a run flag for this behavior makes sense to implement?

joshmarkovic:
**By default**, files in subprojects will be processed multiple times - once for each project in the hierarchy that contains them. For example, a file in `src/backend/` will be checked by hooks in `src/backend/`, then `src/`, then the workspace root.

**To isolate a project**, you can set `orphan: true` in its configuration. When enabled, files in this project are "consumed" by it and will not be processed by parent projects:

```yaml
# src/backend/.pre-commit-config.yaml
orphan: true

repos:
  - repo: https://github.com/astral-sh/ruff-pre-commit
    rev: v0.8.4
    hooks:
      - id: ruff
```

With this option:
- Files in `src/backend/` are processed **only** by hooks in `src/backend/`
- Files in `src/` (but not in `src/backend/`) are processed by hooks in `src/` and the workspace root
- Files in the root (but not in subdirectories with configs) are processed by hooks in the root

This can be useful to avoid redundant processing in monorepos with nested project structures or to completely isolate a subproject from parent configurations.

Closes #1137

### DIFF
--- a/docs/workspace.md
+++ b/docs/workspace.md
@@ -156,6 +156,10 @@ Notice how:
 - The workspace root processes all files in the entire workspace
 - Projects are executed from deepest to shallowest as described in the execution order
 
+#### Orphan Projects and Selectors
+
+When you combine `orphan: true` with selectors such as `--skip`, remember that orphans keep the files they cover. Even if you skip an orphan project (for example via `--skip src/backend/`), that project still claims ownership of the files under its directory. Those files will not fall back to parent projects, so you can disable or precisely target orphaned projects without reintroducing duplicate processing upstream.
+
 ## Command Line Usage
 
 ```bash
--- a/src/cli/run/run.rs
+++ b/src/cli/run/run.rs
@@ -576,10 +576,6 @@ async fn run_hooks(
             .push(hook);
     }
 
-    // Sort projects by their depth in the workspace.
-    let mut project_to_hooks: Vec<_> = project_to_hooks.into_iter().collect();
-    project_to_hooks.sort_by_key(|(_, hooks)| hooks[0].project().idx());
-
     let projects_len = project_to_hooks.len();
     let mut first = true;
     let mut file_modified = false;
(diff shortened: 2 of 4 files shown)


j178:
Ref https://github.com/j178/prek/issues/1137

j178:
Thanks for your effort. Before you spend too much time on this, I think we should talk about the design in #1137. Personally, I feel like this option should be more granular instead of just a global switch.

joshmarkovic:
> Thanks for your effort. Before you spend too much time on this, I think we should talk about the design in #1137. Personally, I feel like this option should be more granular instead of just a global switch.

Exactly how granular should it be? Can you give me more details on the design you have in mind?

On a separate note: I'm new to this repo and some of the tooling used in this project. I don't mind putting in the effort but it might take me some time.

j178:
Thank you!

PR #1133, opened 2025-11-21
Remove completion subcommand from shell completion docs

nblock:
There seems to be no `completion` subcommand the docs for and the other shells don't mention it either.


j178:
Thanks!

PR #1148, opened 2025-11-23
Fix typo “Not” to “Note” in the Automatic Fast Path section

chilin0525:
This PR fixed a typo in `docs/builtin.md` where "Not that" should be "Note that"


j178:
Thanks!

PR #1172, opened 2025-12-02
Support`--cooldown-days` in `prek auto-update`

### LINKED ISSUE #1152: Feature request - cooldown period on autoupdate
I was wondering if prek autoupdate could support a cooldown period (like dependabot) or an absolute cut off date like uv update so you can choose to only install tool updates that have been live for a while (for security reasons).

This is kind of similar / related to https://github.com/j178/prek/issues/174

I'm not sure of the complexity of implementing though. I would think it would only apply to online, e.g. GitHub sourced tools (I guess that's all autoupdate would touch anyway).

lmmx:
- Closes #1152 
- Default cooldown: 1 day, secure by default
- Minimum cooldown: 0 days (no cooldown), making it opt-out
- Maximum cooldown: 255 days (`u8::MAX`), this is more than is typical (Dependabot's is 90 days).
  - No particular handling exists (yet?) for users attempting to put an excessive/invalid number of days

It works :tada:

- I ensured the existing tests will still work by passing cooldown as 0 days.

## Demo

If I rewind the crate-ci/typos hook in the prek repo itself to the last tagged release before [1.40](https://github.com/crate-ci/typos/releases/tag/v1.40.0)

- v1.40.0 of the typos hook was released on 2025-11-26, 5½ days ago from today 2025-12-02

```diff
diff --git a/.pre-commit-config.yaml b/.pre-commit-config.yaml
index 82ee65a..80e9539 100644
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -15,7 +15,7 @@ repos:
         exclude: docs/cli.md # Exclude generated doc
 
   - repo: https://github.com/crate-ci/typos
-    rev: v1.40.0
+    rev: v1.39.0
     hooks:
       - id: typos
```

- 7 day cooldown would mean it had another day to cool off
- 6 day cooldown would mean it would be ready today (at the exact hour)
  - The exact time was 8:26PM GMT, current time is 11:55AM GMT (so it would still be cooling off)
- 5 day cooldown would mean it would have been ready yesterday (so will definitely update)

```sh
louis 🌟 ~/dev/prek $ prek auto-update --cooldown-days 7
[https://github.com/crate-ci/typos] already up to date
[https://github.com/pre-commit/pre-commit-hooks] already up to date
louis 🌟 ~/dev/prek $ prek auto-update --cooldown-days 6
[https://github.com/crate-ci/typos] already up to date
[https://github.com/pre-commit/pre-commit-hooks] already up to date
louis 🌟 ~/dev/prek $ prek auto-update --cooldown-days 5
[https://github.com/crate-ci/typos] updating v1.39.0 -> v1.40.0
[https://github.com/pre-commit/pre-commit-hooks] already up to date
```

### DIFF
--- a/crates/prek/src/config.rs
+++ b/crates/prek/src/config.rs
@@ -628,9 +628,6 @@ pub(crate) struct Config {
     /// When false (default), files in subprojects are processed by both the subproject and
     /// any parent projects that contain them.
     pub orphan: Option<bool>,
-    /// Number of days to wait before updating to a new release.
-    /// Default is 1 day. Set to 0 to disable cooldown.
-    pub cooldown_days: Option<u8>,
 
     #[serde(skip_serializing)]
     #[serde(flatten)]
@@ -941,7 +938,6 @@ mod tests {
                 fail_fast: None,
                 minimum_prek_version: None,
                 orphan: None,
-                cooldown_days: None,
                 _unused_keys: {},
             },
         )
@@ -1018,7 +1014,6 @@ mod tests {
                 fail_fast: None,
                 minimum_prek_version: None,
                 orphan: None,
-                cooldown_days: None,
                 _unused_keys: {},
             },
         )
@@ -1120,7 +1115,6 @@ mod tests {
                 fail_fast: None,
                 minimum_prek_version: None,
                 orphan: None,
-                cooldown_days: None,
                 _unused_keys: {},
             },
         )
@@ -1301,7 +1295,6 @@ mod tests {
                 fail_fast: None,
                 minimum_prek_version: None,
                 orphan: None,
-                cooldown_days: None,
                 _unused_keys: {},
             },
         )
@@ -1439,7 +1432,6 @@ mod tests {
                 fail_fast: None,
                 minimum_prek_version: None,
                 orphan: None,
-                cooldown_days: None,
                 _unused_keys: {},
             },
         )
@@ -1727,7 +1719,6 @@ mod tests {
             fail_fast: None,
             minimum_prek_version: None,
             orphan: None,
-            cooldown_days: None,
             _unused_keys: {},
         }
         "#);
@@ -1815,7 +1806,6 @@ mod tests {
             fail_fast: None,
             minimum_prek_version: None,
             orphan: None,
-            cooldown_days: None,
             _unused_keys: {
                 "local": Object {
                     "language": String("system"),
--- a/crates/prek/src/snapshots/prek__config__tests__read_config.snap
+++ b/crates/prek/src/snapshots/prek__config__tests__read_config.snap
@@ -1,5 +1,5 @@
 ---
-source: src/config.rs
+source: crates/prek/src/config.rs
 expression: config
 ---
 Config {
@@ -286,6 +286,5 @@ Config {
     ),
     minimum_prek_version: None,
     orphan: None,
-    cooldown_days: None,
     _unused_keys: {},
 }


src/cli/auto_update.rs
```diff
@@ -228,11 +235,19 @@ async fn update_repo(repo: &RemoteRepo, bleeding_edge: bool, freeze: bool) -> Re
     let output = cmd.output().await?;
     let mut rev = if output.status.success() {
         let rev = String::from_utf8_lossy(&output.stdout).trim().to_string();
-        let rev = get_best_candidate_tag(tmp_dir.path(), &rev, &repo.rev)
-            .await
-            .unwrap_or(rev);
-        trace!("Using best candidate tag `{rev}`");
-        rev
+        if let Some(best) =
+            get_best_candidate_tag(tmp_dir.path(), &rev, &repo.rev, cooldown_days).await?
```
j178: I think this implementation isn’t quite right. Right now, auto‑update works by grabbing the latest tag that’s closest to `FETCH_HEAD`, and `get_best_candidate_tag` just picks the best tag name if there are multiple tags point to the same commit.

But for cooldown support, we can’t keep using “the latest tag near FETCH_HEAD.” Instead, we should list all the tags, sort them by creation time, and pick the newest one that’s still older than the cooldown period.

lmmx: Hmm, that sounds like my understanding too!

- List all tags
- Filter out any that are not yet ‘cooled down’
- Apply the same distance to pick the ‘best’ for synonymous tags (e.g. when someone tags both ‘latest’ and ‘1.2.3’)

I take it you mean I’ve put the logic that should be outside of / earlier than ‘get best candidate’ inside it, you’re probably right! Sorry

lmmx: so we want to
- List all tags
- Filter out un-cooled tags
- Find the latest from the remaining ones
- If multiple tags point to that tag, pick the best using Levenshtein dist

I kind of conflated the logic here in my focus on getting it to work at all. Will rewrite this way and also break out this mammoth function a bit so it's more readable

(Also to confirm what I think is obvious: bleeding edge should not have a cooldown)

lmmx: Yup, got a failing test repro

```rust
    #[tokio::test]
    async fn test_cooldown_should_fallback_to_older_tag_on_different_commit() {
        let tmp = setup_test_repo().await;
        let repo = tmp.path();

        // Create a commit and tag from 10 days ago
        create_backdated_commit(repo, "first release", 10).await;
        create_tag(repo, "v1.8.0").await;
    
        // Create a commit and tag from 5 days ago
        create_backdated_commit(repo, "second release", 5).await;
        create_tag(repo, "v1.9.0").await;

        // Create a commit and tag from 1 day ago (too new for cooldown)
        create_backdated_commit(repo, "third release", 1).await;
        create_tag(repo, "v2.0.0").await;

        // This is what we WANT to happen - but it will fail with current implementation
        let result = get_best_candidate_tag(repo, "v2.0.0", "v1.8.0", 3).await.unwrap();

        assert_eq!(
            result,
            Some("v1.9.0".to_string()),
            "Should fallback to v1.9.0 (newest tag satisfying cooldown), not None"
        );
    }
```

Which fails:

1. Creates three tags on separate commits at 10, 5, and 1 days ago
2. Calls resolve_revision with a 3-day cooldown
3. Expects v1.9.0 (the newest tag that's at least 3 days old)

The current implementation fails here because `resolve_revision` finds v2.0.0 (closest to HEAD), passes it to `get_best_candidate_tag`, which filters it out for being too new, and returns `None` without then considering v1.9.0 on its different commit!

lmmx: Amended in 25e122b now that test passes.

src/cli/auto_update.rs
```diff
@@ -331,16 +443,93 @@ async fn get_best_candidate_tag(repo: &Path, rev: &str, current_rev: &str) -> Re
...
+    cooldown_days: u8,
+) -> Result<Option<String>> {
+    // Get all version-like tags
+    let stdout = git::git_cmd("git tag")?
+        .arg("tag")
+        .check(true)
+        .current_dir(repo)
+        .output()
+        .await?
+        .stdout;
```
j178: Is it possible to list all tags and their creation dates using a single command, such as `git log --tags --format='%ct %D' --simplify-by-decoration`? This would allow us to avoid repeatedly calling `get_tag_timestamp` in a loop, which I expect would be faster.

lmmx: Awesome! Yeah that looks right.

I think `--simplify-by-decoration` groups by tags, therefore you get a group without a tag.

E.g.

```sh
git log --tags --format='%ct %D' --simplify-by-decoration
```

```
1763719724 tag: chloro-v0.5.3, tag: chloro-core-v0.5.3
1763718760 tag: chloro-v0.5.2, tag: chloro-core-v0.5.2
1763692339 tag: chloro-v0.5.1, tag: chloro-core-v0.5.1
1763679625 
```

I tried and don't think you can dereference the tag to get the `commiterdate` with `git for-each-ref` so you must use `git log`.

<details><summary>Click to show failed demo of for-each-ref</summary>
<p>

I tried this way to avoid that untagged one but it is `taggerdate`, `%ct` is `commiterdate`:

```sh
git for-each-ref --sort=-taggerdate --format='%(refname:strip=2) %(taggerdate:unix)' refs/tags
```

```
chloro-v0.5.3 1763719751
chloro-core-v0.5.3 1763719737
chloro-v0.5.2 1763718788
chloro-core-v0.5.2 1763718773
chloro-v0.5.1 1763692364
chloro-core-v0.5.1 1763692352
```

</p>
</details> 

There is actually nothing after the tag

```sh
louis 🌟 ~/dev/chloro $ git log --tags --format='%ct %D' --simplify-by-decoration | while read line; do echo "1___${line}___2"; done
1___1763719724 tag: chloro-v0.5.3, tag: chloro-core-v0.5.3___2
1___1763718760 tag: chloro-v0.5.2, tag: chloro-core-v0.5.2___2
1___1763692339 tag: chloro-v0.5.1, tag: chloro-core-v0.5.1___2
1___1763679625___2
```

So I would just

- filter the results to the lines with a space (i.e. lines that you can split by " ")
  - then split the part after the space on `", "` to get tags
    - then for each tag strip `"tag: "` prefixes to get the name

lmmx: hm, tests I set up with that failed, perhaps because that isn't what the output looks like.

The reason it doesn't match up is because it doesn't capture tags that are not on a branch, I don't know if that's really going to be a valid scenario (i.e. maybe our test cases just need to ensure the tags are on a branch, real world ones will be)

```sh
louis 🌟 ~/tmp/git_sim $ git log --tags --format='%ct %D' --simplify-by-decoration
1764785355 HEAD -> master, tag: v2.0.0
1764439754 tag: v1.9.0
louis 🌟 ~/tmp/git_sim $ git tag
v1.8.0
v1.9.0
v2.0.0
```

:thinking: 

This works with a subshell ... a batch version of what we already do I guess

```sh
louis 🌟 ~/tmp/git_sim $ git log $(git rev-list --tags) --format='%ct %D'
1764785355 HEAD -> master, tag: v2.0.0
1764439754 tag: v1.9.0
1764007754 tag: v1.8.0
```

The dangling commits in the tests here are unrealistic and if we change them I think it'll work fine with the simplify by decorations flag...

**edit** nope, wasn't that :disappointed:

lmmx: Hmm ok I think it might have been putting single quotes in the arg! You don't need to in here. Anyways, fixed, working! Pushed in 853c728

j178:
I rewrote the `resolve_revision` part and cleaned it up a lot. Also, changing `--cooldown-days` to default to `1` would technically be a breaking change, so I left the default at `0` for now. We can think about switching it to `1` in the next breaking release. I’m also leaning toward staying compatible with pre-commit, but we can figure that out later.

There’s one caveat we should keep in mind: users might expect the “age” to come from the release date (the tag’s creation time). But lightweight tags don’t store any date info. So `prek auto-update` uses the commit date the tag points to instead, which can feel a bit unintuitive.

crates/prek/src/config.rs
```diff
@@ -628,6 +628,9 @@ pub(crate) struct Config {
     /// When false (default), files in subprojects are processed by both the subproject and
     /// any parent projects that contain them.
     pub orphan: Option<bool>,
+    /// Number of days to wait before updating to a new release.
+    /// Default is 1 day. Set to 0 to disable cooldown.
+    pub cooldown_days: Option<u8>,
```
j178, line 633: It looks like `auto-update` doesn’t actually use this config option. I took it out for now — we can always add it back later if we need to.

Issue #785, opened 2025-09-22
Supprt `pyproject.toml` based config


Goldziher:
This issue is an offshot of #593.

I don't think much verbiage is required - basically, it would be great to have `pyproject.toml` support (with or without a special yaml config file). For Python-based projects, it's great. Especially monorepos.

j178:
Could you give me a simple example of what you mean by “pyproject.toml support”? Are you thinking of a local Python hook that uses the local repo (with pyproject.toml) as a dependency?

Goldziher:
No, I mean as an alternative to `.pre-commit-config.yaml`, a central place for configuration.

ulgens:
I wouldn't say `.pre-commit-config.yaml` is the place for configuration, because it seems that configuration options are only available through environment variables. I think it's better to have the repo/hook setup in a `.pre-commit-config.yaml` file but I was expecting to be able to use [these config options](https://prek.j178.dev/configuration/#environment-variables) in `pyproject.toml`

ulgens:
@j178 How does a "needs decision" process works for the project? Do we need to vote somehow? What is the question that needs the decision for?

j178:
If I understand this request correctly, it’s asking for the ability to configure hooks in pyproject.toml under a tool.prek section.

The “needs decision” label basically means it’s waiting for more use cases or discussion before anything gets decided.

ulgens:
> it’s asking for the ability to configure hooks in pyproject.toml under a tool.prek section.

What if we say not hooks but the config options?

j178:
> What if we say not hooks but the config options?

That would be "needs decision" too :)

lmmx:
- Continuing from #593 (that got side-tracked from the topic of YAML!)

> For reference, [here’s a list of reasons](https://github.com/pre-commit/pre-commit/issues/2188) why pre-commit didn’t go TOML. Anthony gives [this example](https://github.com/pre-commit/pre-commit/issues/1165#issuecomment-862603032) of how things would look when using an unchanged structure to represent the same data.

Anthony raises a good point here

> though pre-commit is written in python it is not a python-specific tool -- it does not make sense to add a pyproject.toml file to a non-python project

and in fact, now that prek is written in Rust, I feel like it should be mentioned that this would open up configuring prek in Cargo.toml not just pyproject.toml

[...]

PR #1179, opened 2025-12-02
feat: add pyproject.toml config provider

### LINKED ISSUE #1175: Make env var options configurable via pyproject.toml
Splitting this out from #785 as it is a separate ask:

> I was expecting to be able to use [these config options](https://prek.j178.dev/configuration/#environment-variables) in `pyproject.toml` 

 _Originally posted by ulgens in [#785](https://github.com/j178/prek/issues/785#issuecomment-3340309939)_

The env vars should also be configurable via pyproject.toml settings

---

## Config story

We currently have `src/config.rs` where `.pre-commit-config.yaml` is read in to define both the config schema (the concern of this issue) and the hooks to be provisioned (to be handled separately in #785).

We need to now use it to intercept the env vars we currently directly take from the env in `src/env_vars.rs`, and instead have `config.rs` be our source of truth for env configurable vars (i.e. they can be set in the env but overridden by the config).

- Note that this has historically been a deliberate decision to restrict configuration of prek's predecessor pre-commit, as in issues like [#2696](https://github.com/pre-commit/pre-commit/issues/2696):
  > nope, it can only be set through the command line or an environment variable

This follows the hierarchical control of libraries like [pydantic-settings](https://docs.pydantic.dev/latest/concepts/pydantic_settings/) or [figment](https://docs.rs/figment/latest/figment/) (both these libraries refer to the sources of configuration settings the "providers"

- "Hierarchical" here means that the CLI flags overrule the env vars if both are set

The hierarchy would be:

1. CLI flags
2. `pyproject.toml`
3. `.pre-commit-config.yaml`
4. Env vars

That's the **global** hierarchy though: we're going to start with a separation of config (`pyproject.toml`/env vars) from hook definitions (just `.pre-commit-config.yaml` for now).

We have one config item that can be set by both CLI flags and env vars:

- `--color` = `PREK_COLOR`

That's configured in the clap CLI arg parser itself via the `env` attribute

https://github.com/j178/prek/blob/3ff564e8bfe0e8afb65ff5c427df8cbbb988f94a/src/cli/mod.rs#L144-L152

This gets quite complicated: I am wondering if maybe [figment](https://github.com/SergioBenitez/Figment) would be a good solution for us...

- Test PR to see cargo-bloat result: #1179 

## `prek.toml`

A follow-on PR can introduce `prek.toml`, which would mean we had 5 potential 'providers'.

- I'd let `prek.toml` override `pyproject.toml` like how `.cargo/config.toml` overrides `Cargo.toml`.

lmmx:
Providing config from pyproject.toml as well as env vars and CLI flags

Allows configuration of what were previously env var only settings via pyproject.toml `[tool.prek]` (closes #1175)

- **Note** this does not let you put hook definitions in pyproject.toml, only the env var settings

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -717,8 +717,10 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "8cb01cd46b0cf372153850f4c6c272d9cbea2da513e07538405148f95bd789f3"
 dependencies = [
  "atomic",
+ "parking_lot",
  "pear",
  "serde",
+ "tempfile",
  "toml 0.8.23",
  "uncased",
  "version_check",
@@ -1724,6 +1726,29 @@ name = "parking"
 version = "2.2.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f38d5652c16fde515bb1ecef450ab0f6a219d619a7274976324d5e377f7dceba"
+
+[[package]]
+name = "parking_lot"
+version = "0.12.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "93857453250e3077bd71ff98b6a65ea6621a19bb0f559a85248955ac12c45a1a"
+dependencies = [
+ "lock_api",
+ "parking_lot_core",
+]
+
+[[package]]
+name = "parking_lot_core"
+version = "0.9.12"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "2621685985a2ebf1c516881c026032ac7deafcda1a2c9b7850dc81e3dfcb64c1"
+dependencies = [
+ "cfg-if",
+ "libc",
+ "redox_syscall",
+ "smallvec",
+ "windows-link 0.2.1",
+]
 
 [[package]]
 name = "path-clean"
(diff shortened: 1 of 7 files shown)


lmmx:
What do you make of this approach using figment @j178?

Figment is a settings provider handling tool - [docs](https://docs.rs/figment/latest/figment/)

I put an initial draft of what this would look like in the PR to see the effect on the size  (+1.83% or +0.3MB)

> **edit** I switched to a fork of figment I published myself because the maintainer has left it for a year plus now and it was causing a copy of TOML to be pulled in due to dependency mismatch.

- This would make it much simpler to handle the multiple providers as proposed for #1175 (the issue to make the env vars configurable via pyproject.toml)
- This makes #785 (the issue to make the hook definitions themselves declarable in pyproject.toml/prek.toml) really easy to implement and clear about the hierarchical handling of config overrides

j178:
Haven’t looked too closely yet (work’s been crazy busy), but I’m not really a fan of using such a complicated config framework for the pretty simple configs we use in prek. I might take a closer look later, but it’s probably not going to be a high priority for me.

lmmx:
Ahh ok. I see its benefit as simpler handling of configs from multiple places.

If there's an alternative that you'd prefer to land support for #1175 with let me know.

lmmx:
I've rewritten without figment in that case, I guess we can add it back if we end up with many providers to handle.

Renamed the PR/description accordingly.

j178:
Thank you, @lmmx, as always for your great contribution! Since support of `pyproject.toml` isn't a highly urgent priority right now, I won't be able to get to this immediately, but I really appreciate the work.

lmmx:
Most welcome and not a problem :)

ulgens:
Just wanted to express my support for this feature 🥳

PR #989, opened 2025-10-28
Support Rust language

lmmx:
> - Closes #44

- **chore(wip): initial scaffold for Rust language support**
- **chore(wip): add version parsing like golang's**
- **fix(matches): reject channel is more semantically correct than accept any channel**
- **feat: implement rust installer**
- **feat(deps): install rust deps with cargo add/install**
- ~~**feat(wip): implement run with filesystem path extraction [not correct]**~~
- **feat: reimplement run() using stored version as rustup toolchain specifier**
	- Python stores and uses the original request string (e.g. "stable")
	- Prek stores the resolved semver version (e.g. "1.70.0") after installation
	- Both functionally equivalent with rustup (one uses channel name, other uses specific version number)
- **test: add some tests (working :tada:)**
- **fix: fill in default version for newly downloaded Rust**
- **style(clippy): lint**
- **fix(windows): suppress clippy return type error on overloaded signature**
- **fix(rust): use installed rustup binary path instead of relying on PATH**
- **fix(wip): attempt to pass toolchain (still not set)**
- **fix(rust): query version for cached Rust installations using RUSTUP_TOOLCHAIN**
- **fix: add rustc_bin to path (setup now works in bare env :tada:)**
- **fix(rust): check for rustup at installation location instead of system PATH**
- **test(snapshot): update snapshot for language version test**

## Container testing

<details><summary>Click to show instructions/notes</summary>
<p>

- Build prek first then run container with prek binary mounted
- Use Ubuntu 24.04 as 22.04 has GLIBC that is too old (so would need to target musl linux)

```bash
cargo build && docker run -it --rm -v $(pwd)/target/debug/prek:/usr/local/bin/prek ubuntu:24.04 bash
```

> or to rebuild it in release mode (slower)
> 
> ```bash
> cargo build --release && docker run -it --rm -v $(pwd)/target/release/prek:/usr/local/bin/prek ubuntu:24.04 bash
> ```

Then in container:

```bash
apt-get update && apt-get install -y git curl

mkdir test-project && cd test-project
git init && git config user.email "test@example.com" && git config user.name "Test" && git commit --allow-empty -m "Initial commit"

cat > .pre-commit-config.yaml <<EOF
repos:
  - repo: local
    hooks:
      - id: rust-test
        name: rust-test
        language: rust
        entry: rustc --version
        language_version: stable
        always_run: true
        pass_filenames: false
EOF

git add .
prek run
```

[...]

PR #1180, opened 2025-12-02
rust: add support for finding packages in virtual workspaces

### LINKED ISSUE #1174: `language: rust` fails with virtual workspaces
### Summary

`typos` and `committed` worked with pre-commit until I switched to virtual workspaces (which I did to make python packages to allow my hooks to be installed that way).
- crate-ci/typos#1436
- crate-ci/committed#453

Apparently, this carried forward to prek when it added support for `language: rust`

### Platform

Linux 6.16.3-76061603-generic x86_64 GNU/Linux

### Version

As of #989

### .pre-commit-config.yaml

_No response_

### Log file

_No response_

lmmx:
- Follow-on from #989
- Fixes #1174 (which pre-commit never supported)

## What's new:

- ability to run the `typos-src` hook in the crate-ci/typos repo
- various new unit tests covering different types of workspace including virtual workspaces
- use `cargo_metadata` to get package binary info robustly (+1.22% or +0.2MB binary size)

## Status

- Ready to review (merge if happy)

### DIFF
--- a/src/languages/rust/rust.rs
+++ b/src/languages/rust/rust.rs
@@ -365,11 +365,8 @@ impl LanguageImpl for Rust {
             let toolchain_dir = rustc_bin.parent().expect("Toolchain dir should exist");
             let rustup_home = rustup_home_dir(toolchain_dir);
             vec![
-                (EnvVars::RUSTUP_TOOLCHAIN, toolchain),
-                (
-                    EnvVars::RUSTUP_HOME,
-                    rustup_home.to_string_lossy().to_string(),
-                ),
+                (EnvVars::RUSTUP_TOOLCHAIN, PathBuf::from(toolchain)),
+                (EnvVars::RUSTUP_HOME, rustup_home),
             ]
         } else {
             vec![]
@@ -385,7 +382,7 @@ impl LanguageImpl for Rust {
                 .env(EnvVars::PATH, &new_path)
                 .env(EnvVars::CARGO_HOME, env_dir)
                 .env(EnvVars::RUSTUP_AUTO_INSTALL, "0")
-                .envs(rust_envs.iter().map(|(k, v)| (k, v.as_str())))
+                .envs(rust_envs.iter().cloned())
                 .args(&hook.args)
                 .args(batch)
                 .check(false)


src/languages/rust/rust.rs
```diff
@@ -40,6 +40,81 @@ fn extract_package_name(content: &str) -> Option<String> {
...
+
+/// Extract binary names from `[[bin]]` sections in Cargo.toml.
+fn extract_bin_names(content: &str) -> Vec<String> {
+    let mut names = Vec::new();
+    let mut in_bin_section = false;
+
+    for line in content.lines() {
+        let trimmed = line.trim();
+
+        if trimmed == "[[bin]]" {
```
epage: Would it work to call `cargo metadata --no-deps` (wrapped via [`cargo_metadata` package](https://docs.rs/cargo_metadata/latest/cargo_metadata/))  to discover all packages and their bins?

lmmx: Yep I'm sure it would :slightly_smiling_face: We've tried to avoid adding deps (especially when language-specific) but I can give it a try and see what cargo bloat reports

lmmx: I re-rolled into one commit so we have a 'save point', when I added `cargo_metadata` there's only 1 new cargo lock dep (`cargo-platform`) which I take to mean we already build the rest in the crate tree

```diff
diff --git a/Cargo.lock b/Cargo.lock
index 65ca431..d6311ff 100644
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -368,6 +368,29 @@ dependencies = [
  "serde_core",
 ]
 
+[[package]]
+name = "cargo-platform"
+version = "0.3.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "122ec45a44b270afd1402f351b782c676b173e3c3fb28d86ff7ebfb4d86a4ee4"
+dependencies = [
+ "serde",
+]
+
+[[package]]
+name = "cargo_metadata"
+version = "0.23.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ef987d17b0a113becdd19d3d0022d04d7ef41f9efe4f3fb63ac44ba61df3ade9"
+dependencies = [
+ "camino",
+ "cargo-platform",
+ "semver",
+ "serde",
+ "serde_json",
+ "thiserror 2.0.17",
+]
+
 [[package]]
 name = "cc"
 version = "1.2.47"
@@ -1842,6 +1865,7 @@ dependencies = [
  "async-compression",
  "axoupdater",
  "bstr",
+ "cargo_metadata",
  "clap",
  "clap_complete",
  "ctrlc",
diff --git a/Cargo.toml b/Cargo.toml
index 0d54684..f6347f5 100644
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -97,6 +97,7 @@ tracing-subscriber = { version = "0.3.20", features = ["env-filter"] }
 unicode-width = { version = "0.2.0" }
 which = { version = "8.0.0" }
 glob = "0.3.3"
+cargo_metadata = "0.23.1"
 
 [target.'cfg(unix)'.dependencies]
 prek-pty = { workspace = true }
```

(Just breaking for dinner!)

epage: If bloat is a problem, you can always call `cargo metadata` yourself and define only the set of serde types you need to read the output.  Likely best once you have a working solution with it and if there the need to optimize it is worth it.

lmmx: > +1.22%

I think that's reasonable. This allows us to throw out a lot of processing code (3d3ce36)

j178:
Thanks!

src/languages/rust/rust.rs
```diff
@@ -179,8 +214,28 @@ impl LanguageImpl for Rust {
                     .check(true)
                     .output()
                     .await?;
+            } else if lib_deps.is_empty() {
+                // For workspace members without lib deps, use cargo build + copy
+                // (cargo install doesn't work well with virtual workspaces)
```
epage, line 219: What problem is this? 

(context: also a cargo team member)

lmmx: Here I was trying to install a specific binary from its path whose metadata I got from the workspace, and tried a few ways. I ended up deciding that it’s best to just build and copy the binary rather than try to point to the subpath. I forget exactly what error I was getting but this looked like the best way to do it.

epage: I would assume that the same install approach should be used regardless of whether the package is in the root of a workspace or not.  It keeps the code simpler and makes it faster to discover the root cause of undesired behavior.

The following seemed to work for me
```console
cargo +nightly install -Zunstable-options --dry-run --path ../typos/crates/typos-cli/ --bins
```
(`-Z` is for `--dry-run` to work)

lmmx: Hmmm yes let me throw this in an issue so I can come back to it...

Edit: opened #1191 to track

PR #1182, opened 2025-12-03
Provide more information when validating configs and manifests

MatthewMckee4:
I thought it quite was hard to tell what `validate-config` and `validate-manifest` were doing.

This just provides a bit more information about successes and when the command does nothing (empty paths).

### DIFF
--- a/src/cli/validate.rs
+++ b/src/cli/validate.rs
@@ -1,71 +1,81 @@
 use std::error::Error;
+use std::fmt::Write;
 use std::iter;
 use std::path::PathBuf;
 
 use anstream::eprintln;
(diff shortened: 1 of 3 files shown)


lmmx:
There’s one snapshot causing a test failure ICYMI, you should be able to fix with `cargo insta review` (ignore the ruby/lua/etc ones and just accept the one you changed) or run it specifically with

```sh
cargo insta test --accept unexpected_keys_warning
```

j178:
Thanks! I removed the per‑file success message since it was a bit redundant, and I switched the output to the printer so the message can be suppressed with `-q`.

PR #1183, opened 2025-12-03
Add cache size command

MatthewMckee4:
Very similar to `uv cache size` command.

Also renamed clean to cache clean as I think that is clearer.


j178:
Thanks!

PR #1185, opened 2025-12-03
Fallback to `manual` stage for hooks specified directly in command line

### LINKED ISSUE #1177: Make it easy to "force" running a hook from command line
We have a number of prek hooks that are in hook-stage manual and often we want  to run them with `--all-files`. 

Often when I want to run those hooks, I keep on forgetting to add `--hook-stage manual` - and even if I do, I keep on adding `--all-files`. This usually results in 3 attemps to do it (almost like with inserting an USB-A cable).

It would be great if by default if manual stage hook is used as the only parameter, it would automatically run with both `--hook-stage manual` and `--all-files` - when there is no other hook type to run.

It's generally non-breaking, if you try to do it today, you get:

```
⌁0% [jarek:~/code/airflow] main+ 3m0s 1 ± prek mypy-providers                                
error: No hooks found for stage `pre-commit` after filtering
```

Maybe it's worth to add this special treatment of such manual hooks?

drainpixie:
Closes #1177. I'm aware right now the PR lacks documentation and test changes; I wanted to get a code review first to ensure my approach is correct as it's my first time with this codebase. I have tested the current changes manually as such:

```sh
~/p/o/testing main (nix) $ cat .pre-commit-config.yaml
repos:
  - repo: local
    hooks:
      - id:  foo
        name: Foo 
        entry: echo "foo"
        language: system
        stages: [commit]

      - id: bar 
        name: Bar 
        entry: echo "Bar"
        language: system
        stages: [manual]

      - id: baz
        name: Baz
        entry: echo "Baz"
        language: system
        stages: [manual]

~/p/o/testing main (nix) $ prek    
Foo..................................................(no files to check)Skipped
~/p/o/testing main (nix) $ prek bar
error: No hooks found for stage `pre-commit` after filtering
~/p/o/testing main (nix) $ ../prek/target/debug/prek
Foo..................................................(no files to check)Skipped
~/p/o/testing main (nix) $ ../prek/target/debug/prek bar
warning: No hooks found for stage `pre-commit` after filtering; running with `--hook-stage manual --all-files` as the selected hook(s) support the `manual` stage.
Bar......................................................................Passed
~/p/o/testing main (nix) $ ../prek/target/debug/prek bar baz
warning: No hooks found for stage `pre-commit` after filtering; running with `--hook-stage manual --all-files` as the selected hook(s) support the `manual` stage.
Bar......................................................................Passed
Baz......................................................................Passed
~/p/o/testing main (nix) $ 
```

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -205,7 +205,7 @@ prek run [OPTIONS] [HOOK|PROJECT]...
 </dd><dt id="prek-run--from-ref"><a href="#prek-run--from-ref"><code>--from-ref</code></a>, <code>--source</code>, <code>-s</code> <i>from-ref</i></dt><dd><p>The original ref in a <code>&lt;from_ref&gt;...&lt;to_ref&gt;</code> diff expression. Files changed in this diff will be run through the hooks</p>
 </dd><dt id="prek-run--help"><a href="#prek-run--help"><code>--help</code></a>, <code>-h</code></dt><dd><p>Display the concise help for this command</p>
 </dd><dt id="prek-run--hook-stage"><a href="#prek-run--hook-stage"><code>--hook-stage</code></a> <i>hook-stage</i></dt><dd><p>The stage during which the hook is fired.</p>
-<p>When specified, only hooks configured for that stage (for example <code>manual</code>, <code>pre-commit</code>, or <code>pre-commit</code>) will run. When omitted, prek first attempts to run hooks registered for the <code>pre-commit</code> stage. If no hooks remain after filtering, it automatically falls back to <code>manual</code> so manual-only hooks still execute without having to pass <code>--hook-stage manual</code>.</p>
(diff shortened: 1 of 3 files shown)


potiuk:
I think the interesting cases to add also as manual tests are :

* ~/p/o/testing main (nix) $ ../prek/target/debug/prek foo  -> should work in commit mode
* ~/p/o/testing main (nix) $ ../prek/target/debug/prek foo bar baz (should we only run foo, or foo as commit and bar/bas as manual or fail whole command)

j178:
Thank! I removed the extra fallback to `--all-files` and kept only the manual-stage fallback. `--all-files` controls file selection, which is orthogonal to hook stages, so letting the stage toggle alter file discovery was confusing. Keeping the fallback scoped to `manual` makes the behavior easier to reason about without surprising changes to the file filter logic.

drainpixie:
Glad I could help :)

PR #1197, opened 2025-12-06
Add glob list support for file patterns (`files` and `exclude`)

Goldziher:
This PR adds support for glob patterns:

```yaml
# Top-level defaults
  files:
    glob:
      - src/**/*.rs
      - crates/**/src/**/*.rs
  exclude:
    glob:
      - target/**
      - dist/**

  repos:
    - repo: local
      hooks:
        - id: fmt-rust
          entry: cargo fmt --all
          language: system
          # Single glob is fine too:
          files:
            glob: src/**/*.rs
          # Regex still works as before (default form
  without `glob:`):
          exclude: (^|/)vendor/.*

        - id: lint-js
          entry: pnpm lint
          language: system
          files:
            glob:
              - web/src/**/*.ts
              - web/src/**/*.tsx
          exclude:
            glob: web/src/**/__tests__/**
```

Why? The new monorepo support `prek` offers is nice, but its limited. In glob patterns enhance this further making prek much more versatile - globs are one of the key features of lefthook, that are missing.  Furthermore, the `pattern` usage from pre-commit is inefficient and not very nice to work with - in my humble opinion, so this PR offers a completely backwards compatible alternative. Default is what it always was, while you can opt in to glob patterns.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -125,6 +125,8 @@ This is usually used to narrow down the universe of files in large repositories.
 
     This is a `prek` extension. Upstream `pre-commit` expects regex strings here.
 
+    For more information on the glob syntax, refer to the [globset documentation](https://docs.rs/globset/latest/globset/#syntax).
+
 Examples:
 
 ```yaml
@@ -156,6 +158,8 @@ Global *exclude* regex applied before hook-level filtering.
 !!! note "prek-only globs"
 
     Like `files`, `exclude` supports `glob` (single glob or glob list) as a `prek` extension.
+
+    For more information on the glob syntax, refer to the [globset documentation](https://docs.rs/globset/latest/globset/#syntax).
 
 Examples:
 


Goldziher:
@j178 - hi. Can you enable running github actions for users who are not new to github? its an optional setting. It will make contributing here much simpler.

j178:
Thanks for contributing — this looks good! But since this is a new feature, I’ll need a bit more time to think through the interface design and the overall user experience.

Also, dependencies are normally updated by the Renovate bot, so we usually don’t include them in feature PRs. It doesn’t look like they’re needed for this one, right?

Goldziher:
> Thanks for contributing — this looks good! But since this is a new feature, I’ll need a bit more time to think through the interface design and the overall user experience.
> 
> Also, dependencies are normally updated by the Renovate bot, so we usually don’t include them in feature PRs. It doesn’t look like they’re needed for this one, right?

yup not needed. I just ran cargo update, will undo and rebase.

Goldziher:
@j178 - rebased on main and removed the Cargo.toml/lock version updates. Lemme know if you need me to rebase again in the future.

j178:
Thank you, let's merge!

PR #1221, opened 2025-12-13
Prefer tag creation timestamp in `--cooldown-days`

j178:
Use `git for-each-ref --sort=-creatordate --format='%(refname:lstrip=2) %(creatordate:unix)' refs/tags` to get a list of tags along with their timestamps.
`creatordate` uses the tag’s creation time if it’s an annotated tag, and if it’s a lightweight tag, it falls back to the commit’s timestamp.


potiuk:
There is a very interesting change @j178 -> I am not sure if you realised that - but you can have people asking questions.

I am applying cooldown-days in Airflow in https://github.com/apache/airflow/pull/59395 and I noticed a strange change resulting from `autoupdate --freeze` in prek 0.2.22:

Screenshot 2025-12-14 at 09 26 48

if you look closer, you will see that hash changed for those two repos, but the version did not.

I tried to find the reason and I was surprised that the hash stored previously cannot be found in commits of the repo, yet - surprisingly `git checkout HASH` works (and checks out the new commit hash that prek 0.2.22 updated to).

A bit of checking and it turned out that the original hashes were hashes of the TAGS not hashes of the commits they pointed to  - those tags were signed and previous `auto-update` apparently used hashes of the tags not the commits they pointed out to.

That did not change the security properties of it - but the side effect was that if someone actually moved the tags,  such pre-commit-config would stop working when installing the hook.

So the change in behaviour is "good" -> now the hashes stored are hashes of committs not tags, so even if somoene moves tags, you will see it with `auto-update` but the original pre-commit-config will continue to work.

Just wanted you to know about it in case someone else asks - this behaviour change is not described in release notes (I guess because it was not intentional).

j178:
@potiuk I noticed it, but kinda ignored it haha ^^

This actually came from [this commit](https://github.com/j178/prek/pull/1172/changes/bd66496edf9c19089c7ded4f508399522465dca4#diff-72c6178c38f1611f6ad9493ff27a52d99f8e384c04034ee64b2ea0816d2363c3R361) in #1172, where `freeze_revision` was extracted into its own function. But I accidentally changed `git rev-parse {rev}` to `git rev-parse {rev}^{}`.

According to https://git-scm.com/docs/gitrevisions#Documentation/gitrevisions.txt-revegv0998:

> <rev>^{}, e.g. v0.99.8^{}  
> A suffix ^ followed by an empty brace pair means the object could be a tag, and dereference the tag recursively until a non-tag object is found.

So yeah, this wasn’t intentional. But it also doesn’t really have any practical effect. Whether you use the tag object’s hash or the commit’s hash, both refer to immutable objects in the git history — no one can change them anyway.

And honestly, it might even be slightly better. GitHub shows the underlying commit hash for tags on the releases page, so it’s easier for people to verify that the hash matches what GitHub shows.

j178:
I think I’ll add a test so it doesn’t get changed accidentally again in the future.

potiuk:
> And honestly, it might even be slightly better. GitHub shows the underlying commit hash for tags on the releases page, so it’s easier for people to verify that the hash matches what GitHub shows.

Oh absolutely - it's better this way :) - tag hash is indeed immutable, but the tag itself is not - you can move it even if it is signed and then it will eventually **disappear** after garbage collection. For commit hash you would have to rewrite the history to make the commit disapper, so yeah - referring commit is **definitely** better (though security properties are indeed the same).

PR #1226, opened 2025-12-14
Add a test to make sure `prek auto-update freeze` uses the commit hash when it’s pointed to by an annotated tag

j178:
For https://github.com/j178/prek/pull/1221#issuecomment-3650554198


potiuk (approved):
Nice!

PR #1249, opened 2025-12-20
Implement `check-hook-updates` builtin hook

### LINKED ISSUE #1243: Feature request: `repo: builtin` hook to check for outdated hook versions
## Problem

Remembering to run `prek auto-update` regularly to keep pre-commit hooks up to date is a common challenge. Without regular updates, hooks can become stale, missing important bug fixes, performance improvements, or new features.

## Existing solutions in the pre-commit ecosystem

- **[pre-commit.ci](https://pre-commit.ci/)** — auto-creates weekly PRs with updates (GitHub only)
- **[pre-commit-update](https://github.com/mxr/pre-commit-update)** — a hook that checks for updates during normal pre-commit runs
- **Scheduled CI jobs** — run autoupdate on a cron and open PRs
- **Crontab** — local scheduled updates

## Prek's current state

Prek already has:
- `prek auto-update` command with a nice `--cooldown-days` flag for supply chain security
- `repo: builtin` for fast, zero-setup Rust-native hooks

What's missing is a built-in hook to check/remind developers about outdated hook versions during normal workflow.

## Proposal

Add a `repo: builtin` hook (e.g., `check-hook-updates`) that runs during commits or pushes and warns when hook versions are outdated.

### Example configuration

```yaml
repos:
  - repo: builtin
    hooks:
      - id: check-hook-updates
        # Optional: only warn if updates are older than N days
        # (leveraging the existing --cooldown-days concept)
        args: [--cooldown-days=7]
```

### Behavior

- On commit/push, check if any configured hooks have newer versions available
- If updates are found, print a warning (but don't fail the commit by default)
- Optionally, allow `--fail-on-updates` to make the hook fail if updates are available
- Respect `--cooldown-days` to avoid alerting about very recent releases (supply chain security)

### Why this fits prek's philosophy

This aligns well with prek's "batteries included" approach—providing zero-setup, Rust-native solutions that don't require external dependencies or services. Unlike pre-commit.ci (GitHub-only) or cron jobs (requires system setup), this would work everywhere prek runs, right out of the box.

## Related

- This is different from #1054 which focuses on syncing hook versions with local project dependencies (e.g., matching ruff version in pre-commit config with pyproject.toml). This proposal is specifically about alerting when upstream hooks have newer releases available.
- This is also different from #621 (prek.ci) which discusses a hosted CI service. This proposal is for a local built-in hook that works without any external service.

ichoosetoaccept:
Add a new builtin hook that checks if configured hooks have newer versions available.

## Features
- `--cooldown-days`: minimum release age (in days) for a version to be eligible (default: 0)
- `--fail-on-updates`: fail the hook if updates are available (default: warn only)

## Example usage
```yaml
repos:
  - repo: builtin
    hooks:
      - id: check-hook-updates
        args: [--cooldown-days=3]
```

## Changes
- Added `CheckHookUpdates` variant to `BuiltinHooks` enum
- Implemented hook in `pre_commit_hooks/check_hook_updates.rs`
- Added tests
- Updated documentation
- Added hook to project's own `.pre-commit-config.yaml`

Closes #1243


j178:
Is it possible to eliminate the code duplication from `auto_update.rs`?

ichoosetoaccept:
> Is it possible to eliminate the code duplication from `auto_update.rs`?

Good call. Done!

ichoosetoaccept:
Apologies - my previous commit didn't actually eliminate the duplication as I claimed. I've now properly extracted the shared logic into reusable functions in `auto_update.rs`:

`check_hook_updates.rs` now imports and reuses these instead of duplicating the logic.

crates/prek/tests/builtin_hooks.rs
```diff
@@ -2115,3 +2115,54 @@ fn check_case_conflict_among_new_files() -> Result<()> {
...
+    // The hook should be recognized and run with the arguments
+    cmd_snapshot!(context.filters(), context.run(), @r"
+    success: true
+    exit_code: 0
+    ----- stdout -----
+    check for hook updates...................................................Passed
+
+    ----- stderr -----
+    ");
+}
```
j178, line 2328: I don’t think these two tests are enough, because there aren’t any remote repos in the config file, so the update logic never actually runs. Right?

ichoosetoaccept: You, sir, are correct again.

Just added tests that exercise the update-checking logic with a real remote repo (pre-commit/pre-commit-hooks v4.0.0):

- `check_hook_updates_detects_outdated_repo`: verifies hook runs and detects updates
- `check_hook_updates_fails_on_updates: verifies --fail-on-updates` correctly fails when updates are available

pygarap:
@ichoosetoaccept
1. Can it actually update versions like `prek auto-update`, or does it only check for updates?
2. If you have a `.pre-commit-config.yaml` with a different name and run `prek run --config`, how does this hook behave?

ichoosetoaccept:
@pygarap 
To answer your questions:

1. It **only checks** for updates and reports them—it does not modify the config file. The hook is intentionally read-only: it warns (or fails with --fail-on-updates) when newer versions are available, but leaves the actual updating to prek auto-update.
This design follows the principle that hooks should be predictable and non-destructive by default. Auto-updating during a commit could be surprising and potentially break CI pipelines.
2. It works correctly. The hook accesses its config via hook.project().config(), which is already resolved to the correct config file path when prek run --config <custom-config.yaml> is invoked. The hook doesn't hardcode any config file path—it uses whatever config the parent project was initialized with.

ulgens:
When would this hook run? I tried to understand the "files" arg but couldn't see it in the diff, and I believe it will run everytime prek is triggered. This could cause the hook to run when not needed and generate noise. I wouldn't want this hook to complain about my git hook setup when the only thing I'm doing is editing an unrelated code file.

ichoosetoaccept:
> When would this hook run? I tried to understand the "files" arg but couldn't see it in the diff, and I believe it will run everytime prek is triggered. This could cause the hook to run when not needed and generate noise. I wouldn't want this hook to complain about my git hook setup when the only thing I'm doing is editing an unrelated code file.

@ulgens Great point! I've addressed this by adding a **check interval** feature that throttles how often the hook actually runs.

## Changes

**New default behavior:**
- The hook now only performs the actual update check **at most once every 24 hours** (configurable via `--check-interval-hours`)
- The last check timestamp is stored in the prek cache directory
- Additionally, `--cooldown-days` now defaults to **7 days** (was 0) to avoid suggesting brand-new releases that haven't been vetted by the community yet (supply chain security)

**New arguments:**
- `--check-interval-hours=<N>` (default: 24) — minimum hours between checks. Set to `0` to check every time (useful for CI)
- `--cooldown-days=<N>` (default: 7) — minimum release age before suggesting an update

Does this address your concerns adequately or do you have better ideas/further suggestions?

## Example usage

```yaml
repos:
  - repo: builtin
    hooks:
      - id: check-hook-updates
        # Check at most once per day, only suggest releases older than 7 days
        # (these are the defaults, shown for clarity)
        args: ['--check-interval-hours=24', '--cooldown-days=7']
```

For CI where you want every-run checks:
```yaml
      - id: check-hook-updates
        args: ['--check-interval-hours=0', '--fail-on-updates']
```

## Open question

One thing I noticed: when the check is skipped due to the interval, the hook still shows **"Passed"** rather than **"Skipped"**. This could be slightly misleading since no actual check occurred.

Would it be better to show "Skipped" in this case? I'd need to look into how the hook result reporting works to see if this is feasible without larger changes. Open to suggestions!

ulgens:
> The hook now only performs the actual update check at most once every 24 hours (configurable via --check-interval-hours)

I don't have a strong opposition to this implementation, and one could easily not enable the hook if they don't like it, but it feels a lot like this should be a step/layer in the CI , not a git hook. git hooks are generally (always?) invoked based on the repository state and files, not a pre-defined interval. In contrast, running a weekly task on CI feels much more natural.

ichoosetoaccept:
That's a fair point! You're right that git hooks are traditionally file/state-based, and a time-based interval is unconventional.

A few thoughts on why this might still be useful as a hook:

1. **Discoverability** — Many teams don't have CI set up for every repo, or contributors work on personal forks without CI. Having it as a hook means it "just works" for everyone who clones the repo.

2. **Opt-in flexibility** — Users can:
   - Not enable the hook at all
   - Set `--check-interval-hours=0` for CI-like behavior (check every time)
   - Use `stages: [manual]` to only run on demand (`prek run --hook-stage manual`)

3. **Complementary to CI** — This doesn't replace a weekly CI job; it supplements it by catching outdated hooks during local development before they become a CI failure.

That said, I'm open to alternative approaches. If the time-based behavior feels too unconventional, users can simply set `stages: [manual]` when adding the hook—that way it only runs on demand via `prek run --hook-stage manual`. The hook isn't enabled by default; users choose to add it, so they're already making an explicit decision about when it should run.

j178:
Hey, thanks again for working on this. I’m still thinking through the idea and the details, and I’m not totally sure yet if it’s a good fit as a built‑in hook. I’m going to leave this in draft for now. Sorry for the back‑and‑forth, and thanks for your patience!

joshuagray251-bit:
Thank you to in ok

j178:
Closing, I think we can just use `prek auto-update` directly: https://github.com/j178/prek/issues/1243#issuecomment-4167917063

PR #1256, opened 2025-12-21
Support git urls for rust dependencies

branchv:
This allows specifying git urls in rust's `additional_dependencies`. My motivation is to use `fish_indent` from https://github.com/fish-shell/fish-shell, which isn't published on crates.io. For example:

```yaml
repos:
  - repo: local
    hooks:
      - id: fish_indent
        name: Format fish files
        entry: fish_indent -w
        language: rust
        additional_dependencies:
          - cli:https://github.com/fish-shell/fish-shell:4.2.1
        types: [fish]
```


j178:
Thanks!

Issue #1239, opened 2025-12-19
error: Command `get git root` exited with an error:


jmelahman:
### Summary

I recently migrated my dotfiles to https://www.atlassian.com/git/tutorials/dotfiles which uses a non-default value for `--git-dir`.

I also have a global `~/.githooks` w/ `post-merge`/`post-checkout`, etc. hooks.

Because the `GIT_DIR` is unset in the [hook config](https://github.com/j178/prek/blob/ecf7c5ba2daeb40444e6a0fe33be9cd350931e59/crates/prek/tests/install.rs#L87), all of the subsequent git commands fail (fair enough),

```
$ git --git-dir=$HOME/.dotfiles --worktree=$HOME checkout --
error: Command `get git root` exited with an error:

[status]
exit status: 128

[stderr]
fatal: not a git repository (or any of the parent directories): .git
```

Maybe naively, it seems feasible to instead save the `GIT_DIR` and pass it to prek and use it internally when applicable at least to initialize the hook running.

Repro steps:
```
git clone --bare git@github.com:j178/prek.git /tmp/prek-bare
mkdir /tmp/prek
cd /tmp/prek
git --git-dir=/tmp/prek-bare --work-tree=/tmp/prek checkout --
```

### Platform

Linux 6.17.9-arch1-1 x86_64 GNU/Linux

### Version

0.2.22

### .pre-commit-config.yaml

None/Not relevant

### Log file

$ cat ~/.cache/prek/prek.log 
2025-12-19T08:53:12.489024Z DEBUG prek: 0.2.22
2025-12-19T08:53:12.489048Z DEBUG Args: ["/usr/bin/prek", "hook-impl", "--hook-type=post-merge", "--color=always", "--skip-on-missing-config", "--script-version=3", "--hook-dir", "/home/jamison/.githooks", "--", "0"]
2025-12-19T08:53:12.490315Z TRACE get_root: close time.busy=1.24ms time.idle=2.20µs

PR #1258, opened 2025-12-22
Respect `GIT_DIR` set by git

j178:
Closes https://github.com/j178/prek/issues/1239
Closes https://github.com/j178/prek/pull/1252


j178:
Hey @jmelahman,  
I poked around a bit, and it looks like the right move is to keep `GIT_DIR` and just set the missing `GIT_WORK_TREE`. Can you check if that actually fixed the issue on your end?

jmelahman:
I agree, much cleaner :) I confirmed this resolved my issue. Thank you!

PR #1269, opened 2025-12-23
Unset `GIT_DIR` for auto-update

branchv:
I have the same setup as #1239 for my dotfiles (but using https://yadm.io) and can confirm #1258 fixed it, thank you! The only other issue I've noticed is with auto-update since I set `GIT_DIR=~/.local/share/yadm/repo.git` to use regular git commands and prek gets confused when fetching hook repos:

```console
$ git init /tmp/repo
$ GIT_DIR=/tmp/repo/.git prek auto-update
[https://github.com/crate-ci/typos] update failed: Command `git fetch` exited with an error:

[status]
exit status: 128
```

### DIFF
--- a/crates/prek/src/cli/auto_update.rs
+++ b/crates/prek/src/cli/auto_update.rs
@@ -831,29 +831,4 @@ mod tests {
 
         assert_eq!(rev, Some("v1.2.0".to_string()));
     }
-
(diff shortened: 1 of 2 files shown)


j178:
Thanks, makes sense!

PR #1271, opened 2025-12-23
Support `prek.toml`

### LINKED ISSUE #593: Support `prek.{yaml/toml}` ?
Since I no longer want to use pre-commit due to its author, I would prefer not to see its name even in filenames. Is there any plan to support using a filename like `prek.yaml` even they have similar file structure?

Current not support:

Image

j178:
Closes #593

### DIFF
--- a/crates/prek/src/cli/mod.rs
+++ b/crates/prek/src/cli/mod.rs
@@ -591,7 +591,11 @@ pub(crate) struct ValidateManifestArgs {
 #[expect(clippy::option_option)]
 #[derive(Debug, Args)]
 pub(crate) struct SampleConfigArgs {
-    /// Write the sample config to a file (`.pre-commit-config.yaml` by default).
+    /// Write the sample config to a file.
+    ///
+    /// Defaults to `.pre-commit-config.yaml` unless `--format toml` is set,
+    /// which uses `prek.toml`. If a path is provided without `--format`,
+    /// the format is inferred from the file extension (`.toml` uses TOML).
     #[arg(
         short,
         long,
--- a/crates/prek/src/cli/sample_config.rs
+++ b/crates/prek/src/cli/sample_config.rs
@@ -24,7 +24,7 @@ repos:
 "};
 
 static SAMPLE_CONFIG_TOML: &str = indoc::indoc! {r#"
-#:schema: https://prek.j178.dev/prek.schema.json
+#:schema: https://json.schemastore.org/prek.json
 # Configuration file for `prek`, a git hook framework written in Rust.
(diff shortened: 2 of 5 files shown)


pygarap:
@j178 if `prek.toml` is the equivalent of `.pre-commit-config.yaml`, why not rename it to `prek-config.toml` and also add support for `prek-hooks.toml` as the equivalent of `.pre-commit-hooks.yaml`?

ulgens:
> if prek.toml is the equivalent of .pre-commit-config.yaml, why not rename it to prek-config.toml and also add support for prek-hooks.toml as the equivalent of .pre-commit-hooks.yaml?

Similar question here. Will this be an alternative to `.pre-commit-config.yaml` file or is it to keep prek settings?

OJFord:
@ulgens The description is 'Closes #593', it's the former. (Also as you can see clearly from:
https://github.com/j178/prek/pull/1271/changes#diff-ff219642554f55d0b52b321bd22873805b5075725d97a9caabd22f3e53f0c63bR132)

Regarding the name, my 2p would be that `prek.toml` makes more sense if you forget the history. It's a toml file, obviously it's some kind of config, it doesn't need to be called that the same way I don't have `Cargo-config.toml` or `pyproject-config.toml`. (In fairness, it does seem more popular in the JS world to have `*.config.js`, but I suppose back to my first point - unlike `toml`, `js` isn't readily apparent as a config file.)

ulgens:
@OJFord Oh, thank you. Checking it again now, it seems clear but I was confused at the time.

pygarap:
But what about `prek-hooks.toml`?

crates/prek-consts/src/lib.rs
```diff
@@ -5,9 +5,12 @@ use std::path::Path;
 
 use env_vars::EnvVars;
 
-pub const CONFIG_FILE: &str = ".pre-commit-config.yaml";
-pub const ALT_CONFIG_FILE: &str = ".pre-commit-config.yml";
-pub const MANIFEST_FILE: &str = ".pre-commit-hooks.yaml";
+pub const PRE_COMMIT_CONFIG_YAML: &str = ".pre-commit-config.yaml";
+pub const PRE_COMMIT_CONFIG_YML: &str = ".pre-commit-config.yml";
+pub const PREK_TOML: &str = "prek.toml";
+pub const PRE_COMMIT_HOOKS_YAML: &str = ".pre-commit-hooks.yaml";
```
ulgens: What would be prek equievelant of this file? Or, do we want to remove pre-commit branding from this file later? If so, I worry that how it will play out with `prek.toml`.

PR #1285, opened 2025-12-28
Add `env` to set environment variables for hooks (#1279)

### LINKED ISSUE #1279: Feature request: Ability to set environment variables in `.pre-commit-config.yaml`
It would be really handy to have a `env` or `environment` key to set environment variables.

Right now I have:

```yaml
  - repo: local
    hooks:
      - id: cargo doc
        name: cargo doc
        language: system
        entry: sh -c 'RUSTDOCFLAGS=-Dwarnings cargo doc --all-features --workspace --no-deps --document-private-items' --
        files: \.rs$
        pass_filenames: false
        priority: 0
```

Alternatively, I could implement this with `env`, but I think it would be nicer to just do:

```yaml
  - repo: local
    hooks:
      - id: cargo doc
        name: cargo doc
        language: system
        entry: cargo doc --all-features --workspace --no-deps --document-private-items
        environment:
          RUSTDOCFLAGS: -Dwarnings
        files: \.rs$
        pass_filenames: false
        priority: 0
```

It would be especially useful for customizing hooks from external repos, e.g. if there was an external repo with a `cargo doc` hook, and I wanted to configure `RUSTDOCFLAGS` on it for my local repo.

This would mean that some configurations written for `prek` would not work the same way for `pre-commit`, but that’s already true for workspace support.

This has been requested various times on `pre-commit`, e.g. https://github.com/pre-commit/pre-commit/issues/2367, but perhaps you feel differently? I’m curious about your thoughts on `pre-commit` compatibility (going both ways).

danielparks:
This key is not supported by `pre-commit`.

---

This requires [prek-test-repos/script-hooks](https://github.com/prek-test-repos/script-hooks) to be updated (https://github.com/prek-test-repos/script-hooks/pull/1) before tests will pass. Unfortunately, merging that PR will break tests on this repo until this PR is merged.

I only added a couple of tests here; would you like me to go through and add tests for other languages?

Closes #1279

### DIFF
--- a/crates/prek/tests/languages/script.rs
+++ b/crates/prek/tests/languages/script.rs
@@ -17,36 +17,38 @@ mod unix {
         context.init_project();
         context.write_pre_commit_config(indoc::indoc! {r"
         repos:
-          - repo: https://github.com/danielparks/script-hooks
-            rev: add-env
+          - repo: https://github.com/prek-test-repos/script-hooks
+            rev: v1.0.0
             hooks:
-              - id: echo
+              - id: echo-env
                 env:
-                  WORD2: override
+                  VAR2: universe
+                verbose: true
+              - id: echo-env
+                env:
+                  VAR1: everyone
+                  VAR2: galaxy
                 verbose: true
         "});
         context.git_add(".");
 
-        cmd_snapshot!(context.filters(), context.run(), @r#"
+        cmd_snapshot!(context.filters(), context.run(), @r"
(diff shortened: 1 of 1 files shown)


danielparks:
I added a commit to point the script test at my fork of prek-test-repos/script-hooks; it will need to be removed before merging.

j178:
Thank you, great PR!

PR #1298, opened 2026-01-01
Add `ZhuoZhuoCrayon/throttled-py` to the list of projects using prek

ZhuoZhuoCrayon:
Prek is an excellent project, keep it up! 💪


j178:
Thanks!

PR #1302, opened 2026-01-02
Fix Go template string to detect rootless podman

mocknen:
The current template text (`'{{ .Host.Security.Rootless }}'`) outputs `'true'\n`, so it will never match `true` and cannot detect that podman is in rootless mode. This pull request fixes the issue.


j178:
Thanks!

PR #1306, opened 2026-01-02
Support apple's `container` for docker lang

branchv:
Adding support for https://github.com/apple/container as a container runtime. As a newer project, it's not yet 100% docker-compatible so needed some special casing (missing `--volume`'s  `Z` option and `--init`)

Example:

```console
$ PREK_CONTAINER_RUNTIME=container cargo run -- try-repo https://github.com/gitleaks/gitleaks gitleaks-docker -a
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.14s
     Running `target/debug/prek try-repo 'https://github.com/gitleaks/gitleaks' gitleaks-docker -a`
Using config:
repos:
  - repo: https://github.com/gitleaks/gitleaks
    rev: b66ac75e4fa93d86d78fccd6e2f36d2c0698b2a2
    hooks:
      - id: gitleaks-docker
Detect hardcoded secrets.................................................Passed
```

### DIFF
--- a/crates/prek/src/languages/docker.rs
+++ b/crates/prek/src/languages/docker.rs
@@ -11,7 +11,7 @@ use std::sync::{Arc, LazyLock};
 use anyhow::{Context, Result};
 use lazy_regex::regex;
 use prek_consts::env_vars::EnvVars;
-use tracing::{debug, trace, warn};
+use tracing::{trace, warn};
 
 use crate::cli::reporter::{HookInstallReporter, HookRunReporter};
 use crate::hook::{Hook, InstallInfo, InstalledHook};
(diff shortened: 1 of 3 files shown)


crates/prek/src/languages/docker.rs
```diff
@@ -111,6 +111,7 @@ fn container_id_from_cgroup_v2(mount_info: impl AsRef<Path>) -> Result<String> {
 #[derive(Debug, Copy, Clone, PartialEq, Eq)]
 enum RuntimeKind {
     Auto,
+    Container,
```
j178: This name is a bit confusing. Could we call it something like `apple-container` instead?

branchv: Works for me, I updated this enum to `AppleContainer` and similar elsewhere. But I did keep the user-facing name as `container` for `PREK_CONTAINER_RUNTIME` since it's the name of the executable (and the [package name](https://repology.org/project/container/versions) in homebrew/nix) and needing to use something else there would (at least to me) be pretty counterintuitive / surprising

j178:
Thanks!

Issue #73, opened 2024-11-18
hi


asottile:
the best time for collaboration probably would have been when starting the project but I guess the second best time is now!

the slated long-term plan for pre-commit is to replace the python implementation with rust -- I've been [slowly](https://github.com/pre-commit/pre-commit-rs) working towards that over the last couple of years but I've been busy and progress has been slow.  I figure we can probably accelerate that if you'd like to work together on accomplishing it

there's also the tricky task of compatibility and migration (which I've fortunately thought a lot about and will hopefully make the transition smoother!) -- the main compatibility issues is going to be around regular expressions, `language: pygrep`, and ultimately bootstrapping python where missing for `language: python` (though I think `uv` will likely take that responsibility at some point) -- happy to talk more on this though as it's kinda the hardest part of making such a shift successful

can also help with the planned direction I'm going for with languages in general -- as that's going to break a lot of what isn't yet implemented here (or there) since the structure and internals are likely to change a lot there (goal being to divorce the installations, repositories, and languages as they're really three separate entities currently represented as one (maybe 1.5 if we're being honest))

an aside -- the repository in the current state violates several copyrights and licenses.  I intentionally chose to release my code under MIT to make it very simple to abide by the license (simply distribute a copy of the license in derivative works) but unfortunately your current implementation reads as a plagiarism -- fortunately an easy fix to include the licenses of your copied work!

j178:
I'm thrilled to see you, @asottile!

I apologize for the copyright violation regarding `pre-commit`. I must admit that I essentially *translated* the original Python implementation into this project. I've now added a copyright notice for `pre-commit` to the LICENSE in #74. Thank you so much for your outstanding work on pre-commit!

I started this project purely to practice Rust and was completely unaware of the existence of https://github.com/pre-commit/pre-commit-rs, which has been around for two years! I never intended for this project to become widely known, so I didn't reach out for collaboration—especially since my Rust skills are quite basic and I wasn't sure how far this project would go.

I'm truly honored by your invitation to collaborate! I believe the ideal home for this project is https://github.com/pre-commit/pre-commit-rs. Do you have any specific plans for moving forward? Since I wasn't aware of https://github.com/pre-commit/pre-commit-rs, my implementation took a slightly different path. How do you find the current implementation? Would you prefer a merge, or should we develop anew from the ground up?

[...]

PR #1308, opened 2026-01-03
Add python/cpython to users

ulgens:
https://github.com/python/cpython/pull/143149


ulgens:
It looks a bunch of other Python repos too will be using prek soon:

Screenshot 2026-01-03 at 13 54 34 UTC@2x

j178:
Thanks!

flying-sheep:
I think with this, #73 can be closed

PR #1314, opened 2026-01-06
untar: disallow external symlinks

yihong0618:
as diff

make it safe untar like uv https://github.com/tnotstar/uv/blob/788b70f0fe9565b6d3c674b68afeeaedef6ae9ba/crates/uv-extract/src/stream.rs#L239


j178:
Thanks!

PR #1355, opened 2026-01-14
Remove git env vars from `uv pip install` subprocess

### LINKED ISSUE #1354: GIT_DIR leaks into uv pip install, breaking setuptools_scm file discovery in worktrees
## Summary

When prek auto-installs hook environments during `git commit` in a **git worktree**, the `GIT_DIR` environment variable set by git leaks into the `uv pip install` subprocess. This causes packages using `setuptools_scm` for file discovery to have their `git ls-files` query the wrong repository, resulting in missing package data files.

### Affected scenarios

- **Git worktrees** (100% reproducible) - git sets `GIT_DIR` pointing to the worktree's gitdir
- Any context where `GIT_DIR` or `GIT_WORK_TREE` are set when prek runs

### Affected packages

Any hook package that:
- Uses `setuptools_scm` for file discovery
- Includes non-`.py` package data files (e.g., `.json`, `.yaml`)

Confirmed affected: `validate-pyproject` (missing schema JSON files)

### Reproduction (git worktree)

```bash
# Create a bare repo with worktree
mkdir -p /tmp/worktree-mre/.bare
cd /tmp/worktree-mre/.bare && git init --bare
git worktree add ../main -b main

cd ../main
cat > .pre-commit-config.yaml << 'YAML'
repos:
  - repo: https://github.com/abravalheri/validate-pyproject
    rev: v0.24.1
    hooks:
      - id: validate-pyproject
YAML
cat > pyproject.toml << 'TOML'
[project]
name = "test"
version = "0.1.0"
TOML
git add -A && git commit -m "initial"

# Install prek hook, clean caches
prek install
prek clean
uv cache clean

# Trigger bug via commit
echo "# trigger" >> pyproject.toml
git add pyproject.toml
git commit -m "this will fail"
```

**Expected:** Commit succeeds, validate-pyproject runs

**Actual:**
```
Validate pyproject.toml..................................................Failed
- hook id: validate-pyproject
- exit code: 1

  [ERROR] FileNotFoundError: [Errno 2] No such file or directory:
  '~/.cache/prek/hooks/python-.../validate_pyproject/pyproject_toml.schema.json'
```

### Simulated reproduction (without worktree)

The bug can also be triggered by manually setting `GIT_DIR`:

```bash
mkdir /tmp/mre && cd /tmp/mre
git init
cat > .pre-commit-config.yaml << 'YAML'
repos:
  - repo: https://github.com/abravalheri/validate-pyproject
    rev: v0.24.1
    hooks:
      - id: validate-pyproject
YAML
cat > pyproject.toml << 'TOML'
[project]
name = "test"
version = "0.1.0"
TOML
git add -A && git commit -m "init"

prek clean
uv cache clean

export GIT_DIR="$PWD/.git"
export GIT_WORK_TREE="$PWD"
prek install-hooks

# Verify bug
wc -l ~/.cache/prek/repos/*/src/validate_pyproject.egg-info/SOURCES.txt
# BUG: 44 (should be 239)
```

### Root cause

In a worktree, git sets `GIT_DIR` when running hooks:
```
GIT_DIR=/path/to/.bare/worktrees/main
```

This leaks into `uv pip install`, causing `setuptools_scm`'s `git ls-files` to query the wrong directory. The resulting `SOURCES.txt` has only 44 lines (should be 239) and 0 JSON entries (should be 17).

prek already removes GIT environment variables for its own git commands via `GIT_ENV_TO_REMOVE` in `git.rs`, but this cleanup is not applied when running `uv pip install` in `python.rs`.

### Suggested fix

In `crates/prek/src/languages/python/python.rs`, remove GIT environment variables before running `uv pip install`:

```rust
let pip_install = || {
    let mut cmd = uv.cmd("uv pip", store);
    cmd.arg("pip")
        .arg("install")
        .args(["--project", "/"])
        .env(EnvVars::VIRTUAL_ENV, &info.env_path)
        .env_remove("GIT_DIR")
        .env_remove("GIT_WORK_TREE")
        .env_remove("GIT_INDEX_FILE")
        .check(true);
    cmd
};
```

Or reuse the existing `remove_git_envs()` mechanism.

---

## Platform

Darwin 25.2.0 arm64 (macOS Sequoia)

---

## Version

prek 0.2.28

---

## .pre-commit-config.yaml

```yaml
repos:
  - repo: https://github.com/abravalheri/validate-pyproject
    rev: v0.24.1
    hooks:
      - id: validate-pyproject
```

---

## Log file

The bug is deterministic based on environment variables. Running with `-vvv` shows the build completing without errors, but the resulting package is missing files.

To verify the bug after reproduction:
```bash
wc -l ~/.cache/prek/repos/*/src/validate_pyproject.egg-info/SOURCES.txt
# BUG: 44 (should be 239)

grep -c '\.json$' ~/.cache/prek/repos/*/src/validate_pyproject.egg-info/SOURCES.txt
# BUG: 0 (should be 17)
```

shaanmajid:
## Summary

- Remove GIT environment variables before running `uv pip install` to prevent setuptools_scm file discovery failures
- Add regression test that fails without the fix and passes with the fix

## Problem

When prek runs inside a git worktree, git sets `GIT_DIR` pointing to the worktree's gitdir. This leaks into `uv pip install`, causing packages using `setuptools_scm` for file discovery to have their `git ls-files` query the wrong repository.

This results in missing package data files (e.g., `validate-pyproject`'s schema JSON files).

Fixes #1354

## Test Plan

- [x] `mise run lint` passes
- [x] Verified fix using worktree MRE from issue — `validate-pyproject` hook now installs correctly with all 239 files in SOURCES.txt (was 44)
- [x] Added regression test `git_env_vars_not_leaked_to_pip_install` — verified it fails without the fix and passes with the fix

### DIFF
--- a/crates/prek/tests/languages/python.rs
+++ b/crates/prek/tests/languages/python.rs
@@ -420,3 +420,53 @@ fn pep723_script() -> anyhow::Result<()> {
 
     Ok(())
 }
+
+/// Test that GIT environment variables do not leak into uv pip install subprocess.
+/// When prek runs in a git worktree, git sets `GIT_DIR` which should not propagate to
+/// pip install where it breaks packages using `setuptools_scm` for file discovery.
+///
+/// Regression test for <https://github.com/j178/prek/issues/1354>
+#[test]
+fn git_env_vars_not_leaked_to_pip_install() -> anyhow::Result<()> {
+    let context = TestContext::new();
+    context.init_project();
+
+    // setup.py that fails if GIT_DIR leaks into pip install
+    context
+        .work_dir()
+        .child("setup.py")
+        .write_str(indoc::indoc! {r#"
+        import os, sys
+        from setuptools import setup
+        if os.environ.get("GIT_DIR"):
+            sys.exit("ERROR: GIT_DIR should not leak into pip install")
+        setup(name="test", version="0.1.0", extras_require={"test": []})
+    "#})?;
+
+    context.write_pre_commit_config(indoc::indoc! {r#"
+        repos:
+          - repo: local
+            hooks:
+              - id: check-no-git-dir
+                name: check-no-git-dir
+                language: python
+                entry: python -c "print('ok')"
+                additional_dependencies: [".[test]"]
+                always_run: true
+    "#});
+
+    context.git_add(".");
+
+    // Simulate worktree environment by setting GIT_DIR (like git does in worktrees)
+    cmd_snapshot!(context.filters(), context.run()
+        .env("GIT_DIR", context.work_dir().join(".git")), @r"
+    success: true
+    exit_code: 0
+    ----- stdout -----
+    check-no-git-dir.........................................................Passed
+
+    ----- stderr -----
+    ");
+
+    Ok(())
+}


j178:
Thanks!

PR #1397, opened 2026-01-18
Update deprecated key in renovate config

ulgens:
> matchPaths and matchFiles are now combined into matchFileNames, supporting exact match and glob-only. The "any string match" functionality of matchPaths is now removed

https://github.com/renovatebot/renovate/releases/tag/36.0.0

Supersedes #1396


j178:
Thanks!

PR #1410, opened 2026-01-19
Implement `prek cache gc`

### LINKED ISSUE #31: Implement `prek gc`

j178:
Closes #31

### DIFF
--- a/crates/prek/src/cli/cache_gc.rs
+++ b/crates/prek/src/cli/cache_gc.rs
@@ -26,7 +26,7 @@ pub(crate) async fn cache_gc(
         return Ok(ExitStatus::Success);
     }
 
-    let mut kept_configs = FxHashSet::default();
+    let mut kept_configs: FxHashSet<&Path> = FxHashSet::default();
     let mut used_repo_keys: FxHashSet<String> = FxHashSet::default();
     let mut used_hook_env_dirs: FxHashSet<String> = FxHashSet::default();
     let mut used_tools: FxHashSet<ToolBucket> = FxHashSet::default();
@@ -51,12 +51,12 @@ pub(crate) async fn cache_gc(
                 }
                 err => {
                     warn!(path = %config_path.display(), %err, "Failed to parse config, skipping for GC");
-                    kept_configs.insert(config_path.clone());
+                    kept_configs.insert(config_path);
                     continue;
                 }
             },
         };
-        kept_configs.insert(config_path.clone());
+        kept_configs.insert(config_path);
 
         used_env_keys.extend(hook_env_keys_from_config(store, &config));
 
@@ -91,6 +91,7 @@ pub(crate) async fn cache_gc(
 
     // Update tracking file to drop configs that no longer exist.
     if kept_configs.len() != tracked_configs.len() {
+        let kept_configs = kept_configs.into_iter().map(Path::to_path_buf).collect();
         store.update_tracked_configs(&kept_configs)?;
     }
 


shaanmajid:
Haha, I started looking into the issue and working on an implementation last night, looks like you beat me to it. Your implementation looks better, of course :)

One thing I tried to handle properly was the upgrade experience. Namely, how will existing hook entries installed before gc was implemented be handled?

I was attempting to use the workspace cache files `cache/prek/workspace/*` to derive this; the `tracked_configs` file looks overall better, but it does seem like it may orphan all existing hook entries upon upgrade.

Looking at the flow:
1. User upgrades to a version with cache gc
2. Runs `prek cache gc`
3. `tracked_configs()` returns empty (file doesn't exist yet)
 4. GC exits with "Nothing to clean",  even if there's gigabytes of legitimately orphaned cache

The user would need to run `prek` run on every project first to populate `config-tracking.json` before GC can actually identify unused entries.

Since the workspace cache files already contain `config_files[].path`, you could bootstrap from them when `config-tracking.json` doesn't exist:

```rust
  fn tracked_configs(&self) -> Result<FxHashSet<PathBuf>, Error> {
      let tracking_file = self.config_tracking_file();
      match fs_err::read_to_string(&tracking_file) {
          Ok(content) => /* existing logic */,
          Err(err) if err.kind() == std::io::ErrorKind::NotFound => {
              // One-time migration: bootstrap from workspace caches
              let configs = self.bootstrap_from_workspace_caches();
              if !configs.is_empty() {
                  self.update_tracked_configs(&configs)?;
              }
              Ok(configs)
          }
          Err(err) => Err(err.into()),
      }
  }
```
This gives the best of both worlds: uses `tracked_configs` going forward while seamlessly migrating existing users who already have workspace cache data.

Thoughts?

j178:
I was worried about that too. Bootstrapping from the workspace cache is a great idea, but it also means users have to run `prek cache gc` once before they can use any `prek run`, which might make it less useful.

One option is to release a version with config tracking first, then wait a bit and release another version with `cache gc`. But that might just drag things out longer than necessary.

I’m leaning toward gating `cache gc` behind a `--preview` flag and showing a warning that explains what’s going on. That way, users have to opt into the preview to use cache gc, and they’ll understand why it might not behave exactly as expected.

crates/prek/src/cli/cache_gc.rs
```diff
@@ -0,0 +1,353 @@
...
+}
+
+fn print_removed_details(
+    printer: Printer,
+    verb: &'static str,
+    count: usize,
+    title: &'static str,
+    mut names: Vec<String>,
+) -> Result<()> {
+    names.sort();
```
j178:
```suggestion
    names.sort_unstable();
```

crates/prek/src/cli/cache_gc.rs
```diff
@@ -0,0 +1,353 @@
...
+    store: &Store,
+    dry_run: bool,
+    verbose: bool,
+    printer: Printer,
+) -> Result<ExitStatus> {
+    let _lock = store.lock_async().await?;
+
+    let tracked_configs = store.tracked_configs()?;
+    if tracked_configs.is_empty() {
+        writeln!(printer.stdout(), "Nothing to clean")?;
```
j178:
```suggestion
        writeln!(printer.stdout(), "Nothing to clean".bold())?;
```

shaanmajid:
Great points.

It sounds like there are 3 upgrade paths to consider:

1. An existing user upgrades and immediately runs `prek cache gc`
2. An existing user upgrades and continues working normally (running hooks on various projects). Sometime later they run `prek cache gc`
3. A fresh install user uses `prek` normally for a while, then runs `prek cache gc`

What if we just bootstrap lazily on first GC? Crucially, `prek run` already updates workspace caches during workspace discovery, so during the pre-migration period those caches are effectively tracking configs anyway.

Proposed behavior:
- `prek run`: 
  - If `config-tracking.json` exists, update it.
  - If it doesn't exist AND workspace caches exist, do nothing extra; workspace caches already capture this data.
  - If it doesn't exist AND no workspace caches exist (fresh install), create the tracking file immediately (since there's nothing to migrate).
- `prek cache gc`: If `config-tracking.json` doesn't exist, bootstrap it from workspace caches first, then proceed with GC normally.

This handles all the upgrade scenarios safely:
- Scenario 1: Existing user upgrades -> runs gc immediately -> bootstraps from workspace caches, works correctly
- Scenario 2: Existing user uses prek normally -> runs gc later -> workspace caches accumulated during use, bootstraps, works 
- Scenario 3: Fresh install -> tracking file created on first run -> no migration needed

This way, any required migration always happens transparently on first GC; both new and existing users get correct behavior immediately. No `--preview` flag or no user education necessary :) 

That said, this might be overengineering. If the `--preview` flag approach is simpler to ship and the edge cases are rare enough in practice, that's totally reasonable too. Just wanted to lay out the full picture in case the seamless upgrade path is worth the extra logic.

j178:
Thanks, most of that makes sense. Take a look at #1417. I changed

> If it doesn’t exist AND workspace caches exist, do nothing extra; workspace caches already capture this data.

to

> If it doesn’t exist AND workspace caches exist, bootstrap it from the workspace caches.

Otherwise, `prek run` ends up recording the current config file in `config-tracking.json`, and that breaks the bootstrap step later during `prek cache gc`.

shaanmajid:
@j178 你真牛!

shaanmajid:
> Otherwise, prek run ends up recording the current config file in config-tracking.json, and that breaks the bootstrap step later during prek cache gc.

Sorry, I should have been more clear about:

> If it doesn’t exist AND workspace caches exist, do nothing extra; workspace caches already capture this data.

By "do nothing extra", I meant _don't_ create `config-tracking.json`, and wait until `prek cache gc` runs to bootstrap and create `config-tracking.json` for the first time. I thought it would be better to keep the one-time hit for bootstrapping in `prek cache gc` rather than the first `prek run` after upgrade.

That being said, it's a one-time operation and it should still be pretty fast, so I don't think it matters :)

Nice job on this project, and thanks for incorporating my feedback :)

Issue #619, opened 2025-09-03
Add `language: deno` and `language: bun` support


j178:
Ref #553

janosh:
any way to sponsor this issue? esp. for `deno`?

j178:
I’ll start learning Deno right now! 🤣 (Feel free to sponsor me through GitHub sponsors if you want!)

janosh:
in case it helps for testing, [here's a repo](https://github.com/janosh/matterviz) that already uses `prek` with [`deno lint` and `deno fmt` hooks](https://github.com/janosh/matterviz/blob/01170f667072773e40f59aae7fadc9b8558901d4/.pre-commit-config.yaml#L18-L31). i [manually install `deno` in CI](https://github.com/janosh/matterviz/blob/01170f667072773e40f59aae7fadc9b8558901d4/.github/workflows/lint.yml#L16-L19) for the `language: system` hooks to work there. would be great if `deno` installation and the hook versioning was handled by `prek`

j178:
I spent the whole day trying to get Deno to work with prek, but honestly, I kinda failed. All I know about the JavaScript ecosystem is a bit of Node experience, and Deno is just so different. The way it handles dependencies is really weird to me, and the docs don’t go into enough detail.

For example, I can’t figure out how to expose executable `bin` entrypoints from dependencies, or how to install and reference a locally cloned npm package in Deno. Sorry, but I’m feeling pretty defeated by Deno right now. I’m gonna put this on hold for now until I get more experience or find someone more experienced who can help me out.

For the sponsorship, I’m really sorry - since I probably can’t finish this issue anytime soon, please cancel your sponsorship and I can refund what’s already been paid (by sponsoring you back).

janosh:
> For the sponsorship, I’m really sorry - since I probably can’t finish this issue anytime soon, please cancel your sponsorship and I can refund what’s already been paid (by sponsoring you back).

no need, i use `prek` a lot and enjoy it! keep up the great work and thanks for trying `deno` support 👍 

maybe it helps to mention that `deno` doesn't use `node_modules` by default. dependencies are cached globally and imported via URLs (`deno` does support NPM packages in both `deno.json` and `package.json` via `'npm:'` prefix but not sure that matters for integration with `prek`). maybe i can try to take a closer look myself next week. thanks again

j178:
I put up a pretty rough draft PR here https://github.com/j178/prek/pull/965, if anyone’s interested in checking it out.

[...]

PR #1411, opened 2026-01-19
Add `language: bun` support

shaanmajid:
Adds `language: bun` support for running hooks with the Bun JavaScript runtime.

Mirrors the Node.js implementation:
  - Downloads from GitHub releases (oven-sh/bun)
  - Supports version constraints: `bun@1.x`, `>=1.0, <2.0`, paths
  - Isolated environments with `BUN_INSTALL` redirection
  - `bun install -g` for additional dependencies

Ref #619 (keeping issue open since this PR doesn't implement `deno` support)

### DIFF
--- a/crates/prek/src/languages/bun/version.rs
+++ b/crates/prek/src/languages/bun/version.rs
@@ -40,12 +40,6 @@ impl FromStr for BunVersion {
     }
 }
 
-impl BunVersion {
-    pub(crate) fn new(version: semver::Version) -> Self {
-        Self(version)
-    }
-}
-
 /// `language_version` field of bun can be one of the following:
 /// - `default`: Find system installed bun, or download the latest version.
 /// - `system`: Find system installed bun, or error if not found.
--- a/crates/prek/tests/languages/bun.rs
+++ b/crates/prek/tests/languages/bun.rs
@@ -123,7 +123,7 @@ fn language_version() -> Result<()> {
               - id: bun-version
                 name: bun version check
                 language: bun
-                language_version: "1"
+                language_version: ">1.2"
                 entry: bun -e 'console.log(`Bun ${Bun.version}`)'
                 always_run: true
                 verbose: true
@@ -152,6 +152,7 @@ fn language_version() -> Result<()> {
                 always_run: true
                 verbose: true
                 pass_filenames: false
+                additional_dependencies: ["cowsay"] # different dep to force create separate env
     "#});
 
     context.git_add(".");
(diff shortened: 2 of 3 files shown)


j178:
This looks really good. I went through all the code in one go without any confusion! I’ll take a closer look tomorrow, thank you!

crates/prek/src/languages/bun/installer.rs
```diff
@@ -0,0 +1,292 @@
...
+        let version = versions
+            .into_iter()
+            .find(|version| req.matches(version, None))
+            .context("Version not found on remote")?;
+        Ok(version)
+    }
+
+    /// List all versions of Bun available on GitHub releases.
+    async fn list_remote_versions(&self) -> Result<Vec<BunVersion>> {
+        let url = "https://api.github.com/repos/oven-sh/bun/releases?per_page=100";
```
j178, line 176: I might want to pull the GitHub token from the GH_TOKEN or GITHUB_TOKEN env vars (and document it in configuration.md) so we can avoid hitting the API rate limit.

shaanmajid: Now respects `GITHUB_TOKEN` (automatically set in GitHub actions) for authentication if set.

I did not yet implement support for `GH_TOKEN`, which is set by the GitHub CLI:
1. The `prek` CLI looks like it only uses `GITHUB_TOKEN` so far
2. This codepath should not run so frequently in real-world use to trigger the rate limit; this is primarily a CI concern

Happy to change if you feel otherwise :)

j178: For someone like me living in a country with heavy internet censorship, I have to use a proxy to access GitHub. That means I’m sharing the same busy IP with tons of other people, so hitting the GitHub API rate limit is basically an everyday thing :(

shaanmajid: That's a great point -- and sorry for the situation :(

To be clear though, I did add support for `GITHUB_TOKEN` in [f05643f](https://github.com/j178/prek/pull/1411/commits/f05643f244786ab1dee6aaf8e56daae5264bacad). I'm not sure if I see how supporting `GH_TOKEN` as well may be different than just supporting `GITHUB_TOKEN`. Since this isn't being called by the `gh` CLI in anyway, the user still needs to manually set the environment variable (in e.g. `~/.zshrc`). Am I missing something?

Happy again to add `GH_TOKEN` support as well if you'd prefer it; just want to make sure I fully understand the impact of supporting it.

j178:
Thanks!

PR #1414, opened 2026-01-19
Fix nested list indent rendering issue

mschoettle:
I noticed when reading the documentation that there are a couple bullet points not indented to the second level (in https://prek.j178.dev/#prek-provides-a-better-user-experience):

image

That's because `mkdocs` needs 4 spaces instead of 2.

I ran `mdformat` with the `mdformat-mkdocs` extension and found a couple more spots which are also fixed in this PR.

```shell
uv run --with mdformat --with mdformat-mkdocs mdformat --number **/*.md
```

There were more unrelated changes to this fix which I did not commit to keep the diff small.

NB: Thank you so much for creating `prek`! Love it!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -234,7 +234,7 @@ Big thanks to @lmmx for driving this feature!
 ### Enhancements
 
 - Support`--cooldown-days` in `prek auto-update` ([#1172](https://github.com/j178/prek/pull/1172))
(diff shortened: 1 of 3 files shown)


j178:
Thanks! 😊 Do you also want to take a look at adding mdformat to our .pre-commit-config.yaml so we can catch this kind of thing in the future?

mschoettle:
Done! See #1416

PR #1416, opened 2026-01-19
Add mdformat pre-commit hook

mschoettle:
Follow-up to #1414 to add `mdformat` to the pre-commit config.

* I enabled `--number` to keep numbering of ordered lists in the source
* I enabled `--compact-tables` since the one existing table had the compact form
* I enabled `--align-semantic-breaks-in-lists` to align semantic line breaks in lists to the text as this was already done in existing lists
* The thematic break (horizontal line) has no config option but there is `mdformat-simple-breaks` which I've added (see https://github.com/csala/mdformat-simple-breaks#plugin-rationale)

There are also code formatters available, such as [`mdformat-beautysh`](https://github.com/hukkin/mdformat-beautysh) to format bash code blocks, and [`mdformat-config`](https://github.com/hukkin/mdformat-config) to format YAML code blocks. Those are the main ones being used in this documentation.

Let me know if you want those or if any of the above decisions are made should be changed.

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -12,8 +12,8 @@
 - Rust workspace under `crates/*`. The main CLI binary is `crates/prek` (`src/main.rs`).
 - `prek` is a Rust reimplementation of `pre-commit`: configuration parsing lives in `crates/prek/src/config.rs`, execution/dispatch is in `crates/prek/src/run.rs`, and integration tests exercise the CLI end-to-end under `crates/prek/tests/`.
 - User-facing output is centralized:
-  - Warnings go through `warn_user!` / `warn_user_once!` in `crates/prek/src/warnings.rs` (can be disabled via `-q` / `-qq`).
-  - Progress/output selection is via `Printer` in `crates/prek/src/printer.rs`.
+    - Warnings go through `warn_user!` / `warn_user_once!` in `crates/prek/src/warnings.rs` (can be disabled via `-q` / `-qq`).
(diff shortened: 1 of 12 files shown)


pygarap:
@mschoettle 

[mdformat](https://github.com/hukkin/mdformat) is for pre-commit!
An ultra-fast tool like prek needs to use [rumdl](https://github.com/rvben/rumdl) instead!

[rumdl](https://github.com/rvben/rumdl) is an extremely fast markdown formatter and linter, written from scratch in Rust!

j178:
Thanks! I checked out rumdl a bit and didn’t see an equivalent like `mdformat-mkdocs`, which is the main reason for this PR. I’ll just merge this one for now and maybe migrate later.

rvben:
Hey! Just noticed rumdl was mentioned here. Wanted to clarify that rumdl does actually support MkDocs - you can enable it with:

```toml
[global]
flavor = "mkdocs"
```

This handles admonitions (!!!/???), content tabs (===), snippets, mkdocstrings blocks, auto-references, and the PyMdown extensions like ++inserted++, ==mark==, ^^super^^, etc.

Docs: https://github.com/rvben/rumdl/blob/main/docs/flavors/mkdocs.md

No worries if mdformat works better for your workflow - just wanted to share in case it's useful for future reference!

j178:
@rvben Thanks, I'll take a look!

mschoettle:
@rvben Thanks for chiming in! I had missed that when looking at it briefly. This looks quite interesting. I will have to give `rumdl` a try :)

Issue #1327, opened 2026-01-09
Avoid running `git diff` repeatedly


JelleZijlstra:
### Summary

I was exploring `prek` on a huge work repo (tens of thousands of files) and noticed that it's about 3x slower than pre-commit. Looking at the trace log revealed that `prek` seems to be running `git diff` once per hook. Usually that would be fast but for us it often takes a few seconds, and that adds up when there's a few dozen hooks.

Could you instead run `git diff` once per invocation and reuse the result among hooks?

### Platform

Darwin 25.2.0

### Version

0.2.27

JelleZijlstra:
Dug some more and it looks like `pre-commit` also runs git diff after it ran a hook, but not after a skipped hook (because hooks that run may have changed files, but skipped hooks didn't).

On my repo I was able to improve git diff performance and now prek is comparable in speed to pre-commit, though still a little slower most of the time.

j178:
Thanks! prek runs `git diff` after each hook to check whether any files were modified. There’s probably some room to optimize that, but I don’t really have a better idea right now.

About the skipped hooks: I think that’s a regression from #1232, and it should be fixed in #1335. Could you try that out? If prek is still slow after that, could you share a verbose log (`-vvv` or `~/.cache/prek/prek.log`) so I can take a closer look? Thanks!

yihong0618:
if we use code for `git diff` instead of subcommad is that help tiny performace imporve?

JelleZijlstra:
Thanks! Skipping `git diff` for skipped checks should help a lot, I'll give it another try soon!

PR #1428, opened 2026-01-20
Add integration tests for hook skip behavior

shaanmajid:
Add integration tests verifying that prek correctly identifies and reports skipped hooks in various scenarios.

Tests cover:
- All hooks skipped due to no matching files
- Dry run mode skipping all hooks
- Mixed scenarios with both executed and skipped hooks
- Skipped hooks across different priority groups

Includes regression test for #1335: verifies that `git diff` is not called to check for file modifications when all hooks in a priority group are skipped. Confirmed that test correctly fails without the fix.

Refs: #1327, #1335

### DIFF
--- a/crates/prek/tests/skipped_hooks.rs
+++ b/crates/prek/tests/skipped_hooks.rs
@@ -1,0 +1,220 @@
+//! Integration tests for hook skip behavior.
+//!
+//! These tests verify that prek correctly identifies and reports skipped hooks
+//! in various scenarios: file pattern mismatches, dry-run mode, and mixed
+//! execution across priority groups.
+//!
+//! Includes regression tests for #1335: when all hooks in a group are skipped,
+//! prek should not call `git diff` to check for file modifications.
+
+use anyhow::Result;
+use assert_fs::prelude::*;
+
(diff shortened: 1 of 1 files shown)


crates/prek/tests/skipped_hooks.rs
```diff
@@ -0,0 +1,308 @@
...
+
+    ----- stderr -----
+    "#);
+
+    Ok(())
+}
+
+/// Skipped hooks across different priority groups (serial vs parallel).
+#[test]
+fn skipped_hooks_across_priority_groups() -> Result<()> {
```
j178: I don’t get what this test is about, nothing here seems parallel to me.

shaanmajid: You're right. I tried to actually add parallel hook testing but it didn't go great... it was pretty difficult/flaky to test, so I abandoned it.

I updated this to just test that priority groups work properly.

shaanmajid:
Also consolidated two tests, and considerably simplified the regression test and made it cross-platform (removed `#[cfg(unix)]` gate) by switching to a tracing-based approach instead of the hacky approach I used prior.

j178:
Thanks!

PR #1430, opened 2026-01-20
Add icons to documentation pages

my1e5:
The new Zensical style/format works well when pages have icons attached to them. See, for example, the Zensical docs - https://zensical.org/docs/setup/basics/.

This PR updates the documentation pages to include an `icon:`. I went with what seemed like logical choices for the icons, but please feel free to change them as you see fit.

Before:

Screenshot 2026-01-20 at 21 36 46

After:

Screenshot 2026-01-20 at 21 32 45

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -28,6 +28,7 @@ repos:
         additional_dependencies:
           - mdformat-mkdocs==5.1.3
           - mdformat-simple-breaks==0.1.0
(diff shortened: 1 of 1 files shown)


j178:
I mildly prefer the non‑icon version, it looks clearer to me. I read this article recently (https://tonsky.me/blog/tahoe-icons/), which talks about this icon issue.

image

PR #1433, opened 2026-01-21
Migrate from uv-once-map to mea OnceMap

tisonkun:


j178:
Thank you!

crates/prek/src/languages/python/python.rs
```diff
@@ -86,22 +85,12 @@ pub(crate) async fn query_python_info_cached(
...
-        Err(PythonInfoError::Message(
-            "Python info cache entry missing after wait".to_string(),
-        ))
-    })
+    PYTHON_INFO_CACHE
+        .try_compute(python.clone(), async move || {
+            let info = query_python_info(&python).await?;
+            Ok(Arc::new(info))
+        })
+        .await
```
tisonkun, line 93: Semantic changes a bit. Previously, a failed call would fail all the following calls. I implement `try_compute` so that a failed call wouldn't set the value in the hashmap. Hopefully this is what expected.

j178: That's fine, I think the new behavior makes more sense.

PR #1434, opened 2026-01-21
Bump mea to 0.6.3 to reduce key clone in OnceMap

tisonkun:
This follows #1433 

See https://github.com/fast/mea/pull/109

@j178 toml auto formatted by VS Code. I can revert it if desired, but I suggest we can add a [taplo](https://taplo.tamasfe.dev/) config to keep the toml style consistent. Or even have a pre-commit/prek rule for this.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,11 +17,21 @@ anstream = { version = "0.6.15" }
 anstyle-query = { version = "1.1.5" }
 anyhow = { version = "1.0.86" }
 async-compression = { version = "0.4.18", features = ["gzip", "xz", "tokio"] }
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -43,15 +53,21 @@ libc = { version = "0.2.164" }
 # This is required for the `xz` feature in `async-compression`
 liblzma = { version = "0.4.5", features = ["static"] }
 libyaml = { version = "0.2.0" }
-mea = { version = "0.6.2" }
+mea = { version = "0.6.3" }
```
tisonkun, line 56: The only effective change.

PR #1465, opened 2026-01-22
fix(windows): keep VT enabled during progress

### LINKED ISSUE #1237: Coloring on Powershell broken
### Summary

When installing hooks, the coloring on Powershell is broken. This only happen on install, the coloring on `prek run` is fine.

- It does work with cmd (Version 10.0.26200.7462)
- It doesn't work with builtin Powershell (5.1.26100.7462) or the optional updated one (7.5.4)

Seems to be a Windows only issue, on my mac and linux based systems it works just fine with various shells.

I uploaded a video so you can see it, looks like some ascii encoding issues? First you see that it works on cmd and then the issues in ps https://streamable.com/3h4cni (video will be gone in 2 days)

### Platform

Windows 11 25H2 x86_64

### Version

prek 0.2.22

### .pre-commit-config.yaml

_No response_

### Log file

_No response_

shaanmajid:
Some Windows hook installers (uv, pip, npm, cargo) disable `ENABLE_VIRTUAL_TERMINAL_PROCESSING` on exit, which causes indicatif's progress output to render as raw ANSI escape sequences during `prek install`.

This PR attempts two fixes:
1. Re-enable VT mode after subprocess `output()` and `status()` calls (148fe45df749eda56d8430cd9f067cfdc7002c13)
2. Run a background thread that periodically re-enables VT while progress bars are active (d31b4bdcccf57c1532ae933e9953c5b4b9d4898a)

Notably, the first alone did not fix the issue (see https://github.com/j178/prek/pull/1465#issuecomment-3812945380), but the second did.

Video of issue behavior: https://github.com/user-attachments/assets/10b75c59-f528-4a5e-95cd-c1987ee5de4a
Video of behavior after fix: https://github.com/user-attachments/assets/531000c8-32c7-4d98-a6a9-b1ab05e25582
(both courtesy of @chrisoro)

Fixes #1237

### DIFF
--- a/crates/prek/src/cli/reporter.rs
+++ b/crates/prek/src/cli/reporter.rs
@@ -11,6 +11,53 @@ use crate::hook::Hook;
 use crate::printer::Printer;
 use crate::workspace;
 
+const SPINNER_TICK: Duration = Duration::from_millis(200);
+
+// Windows VT keep-alive to prevent ANSI corruption during subprocess execution.
+//
+// Some Windows tools (uv, pip, npm) disable ENABLE_VIRTUAL_TERMINAL_PROCESSING on exit,
+// causing indicatif's spinner output to render as raw escape sequences. This background
+// thread re-enables VT mode periodically while progress bars are active.
+#[cfg(windows)]
+mod vt_keepalive {
+    use std::sync::Arc;
+    use std::sync::atomic::{AtomicBool, Ordering};
+    use std::thread::{self, JoinHandle};
+
+    pub(super) struct VtKeepAlive {
+        stop: Arc<AtomicBool>,
+        handle: Option<JoinHandle<()>>,
+    }
+
+    impl VtKeepAlive {
+        pub(super) fn new() -> Self {
+            let stop = Arc::new(AtomicBool::new(false));
+            let stop_clone = stop.clone();
+
+            let handle = thread::spawn(move || {
+                while !stop_clone.load(Ordering::Relaxed) {
+                    let _ = anstyle_query::windows::enable_ansi_colors();
+                    thread::sleep(super::SPINNER_TICK);
+                }
+            });
+
+            Self {
+                stop,
+                handle: Some(handle),
+            }
+        }
+    }
+
+    impl Drop for VtKeepAlive {
+        fn drop(&mut self) {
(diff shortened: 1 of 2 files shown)


shaanmajid:
Keeping in draft until @chrisoro or someone with a Windows machine can confirm this actually fixes the issue lolol

chrisoro:
Any chance to build arm64 version of it? I won't have access to my amd64 win machine until end of next week. I do have windows on arm as a vm on my mac

shaanmajid:
Just did a dummy change to `build-binaries.yml` to trigger the build job and build ARM Windows binaries. Those artifacts can be found at https://github.com/j178/prek/actions/runs/21322541711#artifacts.

@chrisoro The link for the ARM Windows standalone binary specifically is https://github.com/j178/prek/actions/runs/21322541711/artifacts/5245733919. Looks like it's a zip file in another zip file lol, but the inner zip has a `prek.exe` which is what you should be able to test with. The rest of the instructions I left in my issue comment should still apply.

Let me know how it goes!

chrisoro:
Tried the arm executable and it shows the same errors in both cmd and powershell. Maybe it is the region/language setting in Windows? Mine is set to Germany / German

shaanmajid:
Sad, can you try the latest build? Tried some more changes.

chrisoro:
Nice, this seems to have done the trick. I can't overserve the broken output anymore. See: https://streamable.com/qaa5qs

shaanmajid:
Awesome! Just to confirm, you were able to consistently reproduce in both cmd and powershell in prek 0.3.0, but the bug is consistently gone in both shells with the fix?

chrisoro:
Correct. I only did it in my Win11 ARM VM and not x86 but that shouldn't matter, right?

shaanmajid:
@j178 no pressure, but not sure if you saw this, it was in draft for a while haha. Any concerns?

j178:
Yeah, I've been keeping an eye on this. Thanks for the hard work! But the solution feels a bit unusual to me. I'm not sure if any other projects run into the same issue or handle it this way. Having a background thread constantly re‑enabling VT just seems a little unfortunate to me :(

j178:
Superseded by https://github.com/j178/prek/pull/2328

PR #1484, opened 2026-01-27
Support `Haskell` language

### LINKED ISSUE #1445: Support `haskell` language

fllesser:
resolve #1445

### DIFF
--- a/crates/prek/src/languages/haskell.rs
+++ b/crates/prek/src/languages/haskell.rs
@@ -1,4 +1,5 @@
 use std::path::Path;
+use std::process::Stdio;
 use std::sync::{Arc, LazyLock};
 
 use anyhow::{Context, Result};
@@ -38,8 +39,8 @@ impl LanguageImpl for Haskell {
 
         debug!(%hook, target = %info.env_path.display(), "Installing Haskell environment");
 
-        let bindir = info.env_path.join("bin");
-        std::fs::create_dir_all(&bindir).context("Failed to create bin directory")?;
+        let bin_dir = info.env_path.join("bin");
+        fs_err::tokio::create_dir_all(&bin_dir).await?;
 
         // Identify packages: *.cabal files in repo + additional_dependencies
         let search_path = hook.repo_path().unwrap_or(hook.project().path());
@@ -65,7 +66,7 @@ impl LanguageImpl for Haskell {
             anyhow::bail!("Expected .cabal files or additional_dependencies");
         }
 
-        // Skip cabal update in CI environment
+        // Run `cabal update` unless explicitly skipped via PREK_INTERNAL__SKIP_CABAL_UPDATE (e.g., in CI)
         if !*SKIP_CABAL_UPDATE {
             // `cabal update` is slow, so only run it once per process.
             CABAL_UPDATE_ONCE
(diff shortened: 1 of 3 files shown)


crates/prek/src/languages/haskell.rs
```diff
@@ -0,0 +1,144 @@
...
+            .arg("--installdir")
+            .arg(&bindir)
+            .args(pkgs)
+            .check(true)
+            .output()
+            .await
+            .context("Failed to install haskell dependencies")?;
+
+        reporter.on_install_complete(progress);
+
```
j178, line 100: We need to call `info.persist_env_path();` here

crates/prek/src/languages/mod.rs
```diff
@@ -47,6 +48,7 @@ static RUBY: ruby::Ruby = ruby::Ruby;
 static RUST: rust::Rust = rust::Rust;
 static SCRIPT: script::Script = script::Script;
 static SWIFT: swift::Swift = swift::Swift;
+static HASKELL: haskell::Haskell = haskell::Haskell;
```
j178: It’d be better to keep the list in alphabetical order.

crates/prek/tests/languages/haskell.rs
```diff
@@ -0,0 +1,138 @@
...
+
+#[test]
+fn remote_hook() {
+    let context = TestContext::new();
+
+    context.init_project();
+
+    context.write_pre_commit_config(indoc::indoc! {r"
+        repos:
+          - repo: https://github.com/fllesser/haskell-hooks
```
j178:
```suggestion
          - repo: https://github.com/prek-test-repos/haskell-hooks
```

It’d be better to keep all the remote hooks under the `prek-test-repos` org. I forked your repo to https://github.com/prek-test-repos/haskell-hooks and added you as a member.

fllesser: I've been waiting for you to fork and fix it, haha.

fllesser:
`cabal update` and `cabal install` seem to be very slow.

To mitigate the race condition issue with `cabal update`, I switched to serialized testing, which now takes over 180 seconds in CI. Are there any solutions to this?

crates/prek/src/languages/haskell.rs
```diff
@@ -0,0 +1,247 @@
...
+        Ok(InstalledHook::Installed {
+            hook,
+            info: Arc::new(info),
+        })
+    }
+
+    async fn check_health(&self, info: &InstallInfo) -> Result<()> {
```
j178: If Haskell spits out a binary and it doesn’t need the compiler to run, then I don’t think we need to check the compiler or Cabal version in check_health.

fllesser: Indeed, modify later

fllesser: Indeed, modify later

j178:
Can we skip `cabal update` in CI via a special env var (e.g `PREK_INTERNAL__SKIP_CABAL_UPDATE`)?

fllesser:
Ok

fllesser:
The haskell-actions/setup is excessively slow，especially macOS

j178:
Thank you!

PR #1490, opened 2026-01-29
Add upgrade suggestion option in config

hofbi:
Currently if the minimum prek version does not match, a user sees something like

```shell
Required minimum prek version `0.3.0` is greater than current version `0.2.30`. Please consider updating prek.
```

It would be nice to give the user a more concrete action to fix this error. To do so, I am adding an optional filed to the config named `prek_upgrade_suggestion`. This allows specifying an extra message like `Please run 'uv tool install prek'` or `Please run prek self upgrade` which is shown on top of the error message above.

### DIFF
--- a/crates/prek/src/config.rs
+++ b/crates/prek/src/config.rs
@@ -1262,7 +1262,8 @@ fn validate_minimum_prek_version(path: &Path, config: &Config) -> Result<(), Err
             .map(str::trim)
             .filter(|suggestion| !suggestion.is_empty())
         {
-            message.push_str(&format!(" Upgrade suggestion: {suggestion}"));
+            use std::fmt::Write as _;
+            let _ = write!(&mut message, " Upgrade suggestion: {suggestion}");
         }
         return Err(Error::InvalidConfig(
             path.user_display().to_string(),
--- a/crates/prek/src/hook.rs
+++ b/crates/prek/src/hook.rs
@@ -973,6 +973,7 @@ mod tests {
                     exclude: None,
                     fail_fast: None,
                     minimum_prek_version: None,
+                    prek_upgrade_suggestion: None,
                     orphan: None,
                     _unused_keys: {},
                 },
--- a/crates/prek/tests/run.rs
+++ b/crates/prek/tests/run.rs
@@ -1926,8 +1926,7 @@ fn minimum_prek_version() {
     ----- stdout -----
 
     ----- stderr -----
-    error: Failed to parse `.pre-commit-config.yaml`
-      caused by: Required minimum prek version `10.0.0` is greater than current version `[CURRENT_VERSION]`. Please consider updating prek.
+    error: Invalid configuration in `.pre-commit-config.yaml`: Required minimum prek version `10.0.0` is greater than current version `[CURRENT_VERSION]`. Please consider updating prek.
     "#);
 }
 
--- a/prek.schema.json
+++ b/prek.schema.json
@@ -133,17 +133,17 @@
       ],
       "default": null
     },
-    "prek_upgrade_suggestion": {
-      "description": "Optional suggestion shown when the minimum prek version is higher than the current version.",
-      "type": [
-        "string",
-        "null"
-      ]
-    },
     "orphan": {
       "description": "Set to true to isolate this project from parent configurations in workspace mode.\nWhen true, files in this project are \"consumed\" by this project and will not be processed\nby parent projects.\nWhen false (default), files in subprojects are processed by both the subproject and\nany parent projects that contain them.",
       "type": [
         "boolean",
+        "null"
+      ]
+    },
+    "prek_upgrade_suggestion": {
+      "description": "Optional suggestion shown when the minimum prek version is higher than the current version.",
+      "type": [
+        "string",
         "null"
       ]
     },


j178:
Thanks for working on this. We’re pretty careful about adding new options, and I don’t think `prek_upgrade_suggestion` on its own is enough reason to create another one. Plus, I don’t think this suggestion really works, since the config writer doesn’t actually know how prek is installed on the user’s side.

hofbi:
Makes sense that you don't want to bloat prek with new options here.

Maybe to clarify the motivation a bit. When using prek on a large project, you typically have a well defined and automated way to setup a developers's environment. With that, we exactly know how developers have installed prek, so being able to give them a concrete command to run would be very useful.

Now if we think about "the config writer doesn’t actually know how prek is installed on the user’s side", would it be possible to get to know this? Because if we would know this somehow, we would not need the `prek_upgrade_suggestion` at all, but prek could automatically suggest the right thing to do.

j178:
> would it be possible to get to know this

We *could* try using a heuristic to detect it — for example, if the current executable is in something like `/opt/homebrew`, we could suggest running `brew upgrade`. But that kind of detection is pretty fragile and only works in a few situations, I’m hesitant to add it.

hofbi:
If you run `uv self update` and uv was not installed by their own install script, they give you a similar vague message, but cannot tell you exactly what to do.

The main motivation here is that if prek is managed by some repository automation, developers might not know how prek was installed. So if the minimum prek version gets increased they have to figure out first how to update prek.

One alternative to adding this to the config would be an environment variable, but not sure if this is an improvement to the current approach 🤔 .

j178:
I tried adding an optional upgrade hint after the version using a semicolon in #1536 , like this:

```yaml
minimum_prek_version: '0.2.0; Use `brew upgrade` to upgrade prek'
```

What do you think?

hofbi:
> I tried adding an optional upgrade hint after the version using a semicolon in #1536 , like this:
> 
> ```yaml
> minimum_prek_version: '0.2.0; Use `brew upgrade` to upgrade prek'
> ```
> 
> What do you think?

@j178  I like it. This option also came to my mind, but I was not sure if I could propose it or if it seems too much of an abuse of the version. So let's do that then. Thanks you!

shaanmajid:
I think we should revert #1536. By modifying the type and behavior of `minimum_prek_version`, that PR is effectively a breaking change.

The purpose of `minimum_prek_version` is to warn users on older versions, but ironically, the new syntax crashes on any prek version before #1536 (`<= 0.3.1`), even when the version requirement is met.

**MRE:**

```yaml
minimum_prek_version: '0.2.0; Use brew upgrade to upgrade prek'
repos:
  - repo: local
    hooks:
      - id: test
        name: test
        entry: echo hi
        language: system
```

```console
$ uvx prek@0.3.0 run -a
error: Failed to parse `.pre-commit-config.yaml`
  caused by: unexpected character ';' after patch version number
```

Note that prek 0.3.0 is *greater* than `minimum_prek_version: '0.2.0'`, yet it still errors. The feature defeats itself - old versions can't parse the new syntax, so they never reach the code that would display the upgrade hint.

In summary:
1. Doesn't help users who need to upgrade (they crash before seeing the hint)
2. Actively breaks users who don't need to upgrade (they meet the requirement but can't parse the syntax)

More broadly, I agree with @j178's earlier point:

> I don't think this suggestion really works, since the config writer doesn't actually know how prek is installed on the user's side.

---

@hofbi I do agree this is a meaningful issue; I think we all can make the experience a bit better:

**Individual projects:**
- Use project/environment managers (e.g. `mise`, `uv`, etc.) to enforce a consistent development environment and prek version for contributors
- Document setup instructions in CONTRIBUTING.md
- Enforce a stable prek version in CI (for example, using [prek-action](https://github.com/j178/prek-action))

**Hook maintainers:**
- Set `minimum_prek_version` in hook manifests when relying on specific prek features or behavior changes (though this is rare since most hooks target upstream pre-commit)

**prek itself:**
- As mentioned earlier, detect the install method and suggest the correct upgrade command (but also, as mentioned earlier, this can be difficult...)

uv, for example, [detects Homebrew installations](https://github.com/astral-sh/uv/blob/ee6e3be815e4d4b18667e521aa3faba4321446cb/crates/uv/src/install_source.rs) by inspecting the executable path for `/Cellar/uv/`, and suggests the appropriate upgrade command.  Notably, they _only_ do this for Homebrew for now, but it is something we can implement:

```console
$ uv self update
error: uv was installed through an external package manager and cannot update itself.

hint: You installed uv using Homebrew. To update uv, run `brew update && brew upgrade uv`
```

prek could take a similar approach. When `prek self update` fails (as it does today for managed installs):

```console
$ prek self update
error: prek was installed through an external package manager, and self-update is not available.
       Please use your package manager to update prek.
```

We could enhance this to detect Homebrew and suggest `brew upgrade prek`. Similarly, when `prek run` fails due to `minimum_prek_version`:

```console
$ prek run -a
error: Required minimum prek version `0.4.0` is greater than current version `0.3.1`.
       Please consider updating prek.
```

We could further suggest the appropriate upgrade command based on detected install method.

We could start conservatively with path heuristics (e.g., `/Cellar/prek/` for Homebrew, `/.cargo/bin/` for cargo) and expand over time with build-time feature flags for official packages or marker files / flags we can write during environment-specific installation steps. But even minimal detection would be more helpful than asking config authors to guess.

shaanmajid:
I will try to get started on a PR for Homebrew and cargo detection...

PR #1492, opened 2026-01-29
fix(node,bun): include toolchain in PATH during install

shaanmajid:
## Summary

PR #1488 removed symlinks from `bin_dir` but only updated `run()` to prepend the toolchain directory to PATH. The `install()` functions still only prepended `bin_dir`, causing `npm install` to fail when no system node is available in PATH.

## MRE

With `.pre-commit-config.yaml`:
```yaml
repos:
  - repo: https://github.com/pre-commit/mirrors-prettier
    rev: v3.1.0
    hooks:
      - id: prettier
        language_version: "24.13.0"
```

When no system node is in PATH (can be simulated with `PATH=/usr/bin:/bin`, running:
```bash
prek run --all-files
```

Fails with
```
error: Failed to install hook `prettier`
  caused by: Command `npm install` exited with an error:
env: node: No such file or directory
```

With system node in PATH, `npm install` "works" but uses wrong node version (system instead of requested).

## Fix

Mirror the `run()` PATH setup in `install()` by including the toolchain's parent directory:

```rust
let node_bin = node.node().parent().expect("Node binary must have parent");
let new_path = prepend_paths(&[&bin_dir, node_bin])?;
```

Note: Bun fix is for consistency/edge cases (bun is a native binary, not a shebang script like npm).

### DIFF
--- a/crates/prek/tests/common/mod.rs
+++ b/crates/prek/tests/common/mod.rs
@@ -421,9 +421,8 @@ macro_rules! cmd_snapshot {
 #[allow(unused_imports)]
 pub(crate) use cmd_snapshot;
 
-#[allow(clippy::disallowed_methods)]
 pub(crate) fn remove_bin_from_path(bin: &str, path: Option<OsString>) -> anyhow::Result<OsString> {
-    let path = path.unwrap_or(std::env::var_os("PATH").expect("PATH not set"));
+    let path = path.unwrap_or(EnvVars::var_os(EnvVars::PATH).expect("Path must be set"));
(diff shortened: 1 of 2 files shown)


shaanmajid:
Note: The existing `doctoc` test (`node.rs:201`) would have caught this regression, but it's ignored for being slow/flaky (fetches from GitHub).

Added `additional_dependencies_without_system_node` as a fast, local regression test that exercises the same code path.

j178:
Thanks!

PR #1494, opened 2026-01-29
Add GitHub artifact attestations to release workflow

### LINKED ISSUE #1493: Add GitHub artifact attestations to release workflow
## Summary

prek publishes release artifacts via `cargo-dist` in `.github/workflows/release.yml` but does not generate GitHub artifact attestations for those assets. This means consumers can verify checksums, but cannot verify provenance (that the artifact was built by the official release workflow).

## Proposal
Add an “Attest” step to the release workflow’s announce job. After adding attestations, consumers should be able to run:
```bash
gh attestation verify <artifact> --repo j178/prek
```
to see a verified provenance statement.

shaanmajid:
Closes #1493

Adds [actions/attest-build-provenance](https://github.com/actions/attest-build-provenance) to the release workflow so consumers can verify artifacts were built by CI:

```bash
gh attestation verify <artifact> --repo j178/prek
```

GitHub already generates release attestations that verify file integrity (checksums), but build provenance additionally proves the artifacts were produced by the CI workflow rather than uploaded manually.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -277,6 +277,10 @@ jobs:
     # "host" however must run to completion, no skipping allowed!
     if: ${{ always() && needs.host.result == 'success' && (needs.publish-homebrew-formula.result == 'skipped' || needs.publish-homebrew-formula.result == 'success') && (needs.custom-publish.result == 'skipped' || needs.custom-publish.result == 'success') }}
     runs-on: "ubuntu-latest"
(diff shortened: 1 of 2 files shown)


j178:
Thanks! I think it would also be great to add attestation for our `build-docker.yml` as well.

shaanmajid:
Ah, I missed that. I'll make a follow up PR :)

PR #1497, opened 2026-01-30
Add GitHub artifact attestations to Docker workflow

shaanmajid:
Adds [actions/attest-build-provenance](https://github.com/actions/attest-build-provenance) to the Docker build workflow so consumers can verify images were built by CI:

```bash
gh attestation verify oci://ghcr.io/j178/prek:latest --repo j178/prek
```

Similar to #1494 which added attestations for release artifacts.

Reference: [astral-sh/uv#8685](https://github.com/astral-sh/uv/pull/8685)

PR #1500, opened 2026-01-30
docs: add commands for artifact verification using GitHub Attestations

shaanmajid:
Document how to verify release artifacts and Docker images using `gh attestation verify`.

Refs: #1494, #1497


j178:
Thank you!

PR #1501, opened 2026-01-30
docs: remove `todo.md`

shaanmajid:
- Remove `todo.md` as remaining items have been addressed or documented elsewhere.
- Move note to `diff.md` that `pre-commit hazmat` subcommand (v4.5.0) is not implemented.


j178:
There are still a few references to `todo.md` (you can find them by searching for `/todo`). One of them shows up in the “unimplemented warning.” If we remove the todo page, I think we’ll need to redirect the `/todo` URL to a new page instead of just returning a 404. Is that something MkDocs can do?

shaanmajid:
Gah, thanks for pointing that out.
Can't redirect I don't think, unfortunately :-( There _is_ [mkdocs-redirect](https://github.com/mkdocs/mkdocs-redirects), but we can't use that because Zensical doesn't support plugins...

~I'd be willing to cleanup all references if you're onboard. Alternatively, we can just keep todo.md around as a file, but remove from nav for now, and create a tracking issue for eventual long-term cleanup.~

Actually, looking at all the references, I noticed they all seemed to be about language support, so I just updated the wording and links to direct to the language support section.

j178:
Thanks!

PR #1520, opened 2026-02-01
Replace serde-yaml with serde-saphyr (again)

j178:
serde-saphyr has way better error message, and it is written in safe Rust.

Previous try: #1087
Previous revert: #1106 and #1112

TODO:
Replace serialize_yaml_scalar with serde-saphyr as well:

https://github.com/j178/prek/blob/0a4be22599076b8c6273940b2760828b442c78f4/crates/prek/src/yaml.rs#L12-L13


crates/prek/src/config.rs
```diff
@@ -1517,8 +1529,18 @@ mod tests {
...
-        let result = serde_yaml::from_str::<Config>(yaml);
-        insta::assert_snapshot!(result.unwrap_err().to_string(), @"repos: Invalid remote repo: missing field `id` at line 2 column 3");
+        let result = serde_saphyr::from_str::<Config>(yaml);
+        insta::assert_snapshot!(result.unwrap_err().to_string(), @"
+        error: line 2 column 5: Invalid remote repo: missing field `id` at line 2, column 5
+         --> <input>:2:5
+          |
+        1 | repos:
+        2 |   - repo: https://github.com/crate-ci/typos
+          |     ^ Invalid remote repo: missing field `id` at line 2, column 5
```
j178, line 1539: This should be at the line 4

PR #1525, opened 2026-02-02
fix(ci): run zizmor on push to master, not main

shaanmajid:


yihong0618:
nice catch!

shaanmajid:
@yihong0618 谢谢朋友 :)

PR #1527, opened 2026-02-02
ci: tidy up workflows and restore lost bits

shaanmajid:
- Add missing `test-code` gate to `ecosystem-cpython` job
- Add timeout to `check-release` for consistency  
- Gate `setup-mold` to Linux-only where applicable
  - Mold only supports Linux; macOS support was removed in v1.8.0 (moved to commercial "sold" product)
- Use `run-bench` for performance job (more precise gating)
- Restore swift tests on Ubuntu (preinstalled on runners!)
- Bring back helpful comments lost in language test split

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -111,6 +111,7 @@ jobs:
   check-release:
     name: "check release"
     needs: plan
+    timeout-minutes: 10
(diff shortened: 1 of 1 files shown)


j178:
Thanks!

PR #1538, opened 2026-02-02
Revert "Allow optional upgrade hints after semicolon in `minimum_prek_version`"

shaanmajid:
Reverts #1536. This PR was inadvertently a breaking change and causes errors when the hint syntax is encountered in versions <= 0.3.1.

See #1490 discussion and [MRE](https://github.com/j178/prek/pull/1490#issuecomment-3837782298) for more info.


j178 (approved):
Thanks for the detailed breakdown. it really is a breaking change. I honestly didn’t think of that.

PR #1539, opened 2026-02-03
fix(tests): use version filter in minimum_prek_version snapshots

shaanmajid:
Replace hardcoded version `0.3.1` with `[VERSION]` placeholder using insta's filter mechanism. This prevents test failures when the version is bumped.

### DIFF
--- a/crates/prek/src/config.rs
+++ b/crates/prek/src/config.rs
@@ -1172,6 +1172,12 @@ mod tests {
     use super::*;
     use std::io::Write as _;
 
+    /// Filter to replace dynamic version in snapshots
+    const VERSION_FILTER: (&str, &str) = (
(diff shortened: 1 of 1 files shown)


crates/prek/src/config.rs
```diff
@@ -1677,16 +1685,18 @@ mod tests {
...
-        3 |   entry: echo test
-          |
-        ");
+        insta::with_settings!({ filters => vec![VERSION_FILTER] }, {
+            insta::assert_snapshot!(err, @"
+            error: line 1 column 3: Required minimum prek version `10.0.0` is greater than current version `[CURRENT_VERSION]`. Please consider updating prek. at line 1, column 3
+             --> <input>:1:3
+              |
+            1 | - id: test-hook
+              |   ^ Required minimum prek version `10.0.0` is greater than current version `[CURRENT_VERSION]`. Please consider updating prek. at line 1, column 3
```
j178, line 1694: This error location isn't right, I'll take a look.

Issue #1438, opened 2026-01-21
Reduce amount of repeated logs


OttoAndrey:
### Summary

Hi! Thank you for useful package!

I have a question about logs. For github action I see many repeated logs about "instaling" hooks. Is it possible to reduce their amount? Maybe by some setting? Locally it looks ok.
I tried but didn't find similar issues or settings im docs

<img width="397" height="664" alt="Image" src="https://github.com/user-attachments/assets/44c190cf-2828-4cbe-976f-cac367a08e32" />

For example pre-commit for same yaml file looks ok
<img width="640" height="488" alt="Image" src="https://github.com/user-attachments/assets/61139e6e-966d-4add-ba02-7fa9c17300ab" />

### Platform

Ubuntu

### Version

0.2.30

j178:
This is an “Installing hooks” progress bar. It shouldn’t show up in GitHub Actions, since stderr shouldn’t be treated like a real TTY there. I’ll take a look.

OttoAndrey:
@j178 thank you! for more context: i ran it by poetry. mb its important to now for debugging

<img width="639" height="702" alt="Image" src="https://github.com/user-attachments/assets/88dec47a-9dce-4c5b-adc3-37b80bab1eca" />

yihong0618:
its kind of hard problem

another things check for  tty and in ci 


help with claude I wrote a script to reproduce this



```sh
#!/bin/bash
set -e

SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
PROJECT_ROOT="$(cd "${SCRIPT_DIR}/.." && pwd)"

PREK="$PROJECT_ROOT/target/release/prek"

TEMP_DIR=$(mktemp -d)
trap "rm -rf $TEMP_DIR" EXIT

cd "$TEMP_DIR"

echo "Setting up test project..."
echo ""

# Initialize git repo
git init -q
git config user.name "Test"
git config user.email "test@test.com"

# Create pre-commit config
cat > .pre-commit-config.yaml << 'EOF'
repos:
  - repo: https://github.com/pre-commit/pre-commit-hooks
    rev: v5.0.0
    hooks:
      - id: trailing-whitespace
      - id: end-of-file-fixer
      - id: check-yaml
EOF

# Create a test file with issues
mkdir -p src
echo "test   " > src/sample.txt
git add .

# Create a Python script that simulates poetry's PTY behavior in GitHub Actions
[...]

PR #1552, opened 2026-02-03
fix: hide progress bars in pseudo-TTY environments

### LINKED ISSUE #1551: Garbled stdout (spinner/control characters) when running prek in Neovim/Fugitive
### Summary

On my Neovim setup I use Fugitive to commit, which runs my `pre-commit` hooks. When I run `pre-commit` directly, the stdout renders correctly. But when it runs via `prek` , the output includes strange/garbled characters (looks like control sequences from a progress spinner).

Is this expected? If not, is there a way to disable the spinner/progress animation so the output is plain text?

Environment:

* Neovim
* vim-fugitive (commit workflow)
* `pre-commit` hooks run during commit

### Platform

Darwin 24.6.0 arm64

### Version

0.2.29

### .pre-commit-config.yaml

```yaml
default_stages:
  - pre-commit

default_install_hook_types:
  - pre-commit
  - commit-msg
  - pre-push

repos:
  - repo: local
    hooks:
      - id: ruff-format
        name: ruff format
        entry: uv run ruff format --verbose
        language: system
        types_or: [python, pyi]

      - id: ruff-check
        name: ruff check
        entry: uv run ruff check --fix --config=pyproject.toml
        language: system
        types_or: [python, pyi]

      - id: ty-check
        name: ty check
        entry: uv run ty check src/
        language: system
        types_or: [python, pyi]

  - repo: builtin
    hooks:
      - id: trailing-whitespace
      - id: check-added-large-files
      - id: end-of-file-fixer
      - id: detect-private-key
      - id: trailing-whitespace
      - id: check-added-large-files
      - id: check-case-conflict
      - id: end-of-file-fixer
      - id: fix-byte-order-marker
      - id: check-json
      - id: check-toml
      - id: check-yaml
      - id: check-xml
      - id: mixed-line-ending
      - id: check-symlinks
      - id: check-merge-conflict
      - id: detect-private-key
      - id: check-executables-have-shebangs

  - repo: https://github.com/commitizen-tools/commitizen
    rev: v4.10.0
    hooks:
      - id: commitizen
        stages:
          - commit-msg

  - repo: local
    hooks:
      - id: pytest
        name: pytest (quick)
        entry: uv run pytest -q -m "not slow" --maxfail=1 --disable-warnings
        language: system
        types: [python]
        pass_filenames: false
        stages: [pre-push]


  - repo: https://github.com/astral-sh/uv-pre-commit
    # uv version.
    rev: 0.9.4
    hooks:
      - id: uv-lock
```

### Log file

Here is an example of the render output that I see on neovim.
```text
[2Kruff format..............................................................
^[[2K^[[1Aruff format..............................................................                   ^[[1A
^[[2Kruff format..............................................................Passed
^[[2Kruff check...............................................................
^[[2K^[[1Aruff check...............................................................                   ^[[1A
^[[2Kruff check...............................................................Passed
^[[2Kty check.................................................................
^[[2K^[[1Aty check.................................................................                   ^[[1A
^[[2K^[[1Aty check.................................................................                   ^[[1A
^[[2Kty check.................................................................Passed
^[[2Kdocstring coverage.......................................................
^[[2K^[[1Adocstring coverage.......................................................                   ^[[1A
^[[2Kdocstring coverage.......................................................Failed
- hook id: docstring-coverage
- exit code: 2

  error: Failed to spawn: `docstr-coverage`
    Caused by: No such file or directory (os error 2)
^[[2Ktrim trailing whitespace.................................................Passed
^[[2Kcheck for added large files..............................................Passed
^[[2Kfix end of files.........................................................Passed
^[[2Kcheck for case conflicts.................................................Passed
^[[2Kfix utf-8 byte order marker..............................................Passed
^[[2Kcheck json...........................................(no files to check)Skipped
^[[2Kcheck toml...........................................(no files to check)Skipped
^[[2Kcheck yaml...........................................(no files to check)Skipped
^[[2Kcheck xml............................................(no files to check)Skipped
^[[2Kmixed line ending........................................................Passed
^[[2Kcheck for broken symlinks............................(no files to check)Skipped
^[[2Kcheck for merge conflicts................................................Passed
^[[2Kdetect private key.......................................................Passed
^[[2Kdon't commit to branch...................................................Passed
^[[2Kcheck that executables have shebangs.................(no files to check)Skipped
^[[2Kuv-lock..............................................(no files to check)Skipped
^[[2K
```

shaanmajid:
Fixes #1551

Progress bars now respect anstream's TTY detection instead of indicatif's independent check. This prevents garbled ANSI escape codes (`^[[2K`, `^[[1A`) in environments like Neovim/Fugitive which use pseudo-TTYs, but don't interpret cursor movement codes.

### DIFF
--- a/crates/prek/src/printer.rs
+++ b/crates/prek/src/printer.rs
@@ -43,7 +43,13 @@ impl Printer {
         match self {
             Self::Silent => ProgressDrawTarget::hidden(),
             Self::Quiet => ProgressDrawTarget::hidden(),
-            Self::Default => ProgressDrawTarget::stderr(),
+            Self::Default => {
+                if crate::run::supports_progress() {
+                    ProgressDrawTarget::stderr()
+                } else {
+                    ProgressDrawTarget::hidden()
+                }
+            }
             // Confusingly, hide the progress bar when in verbose mode.
             // Otherwise, it gets interleaved with debug messages.
             Self::Verbose => ProgressDrawTarget::hidden(),
--- a/crates/prek/src/run.rs
+++ b/crates/prek/src/run.rs
@@ -1,5 +1,6 @@
 use std::cmp::max;
 use std::ffi::OsStr;
+use std::io::IsTerminal;
 use std::path::Path;
 use std::sync::LazyLock;
 
@@ -18,6 +19,25 @@ pub(crate) static USE_COLOR: LazyLock<bool> =
         // We just asked anstream for a choice, that can't be auto
         ColorChoice::Auto => unreachable!(),
     });
+
+fn term_is_dumb() -> bool {
+    EnvVars::var_os(EnvVars::TERM)
+        .and_then(|term| term.to_str().map(|term| term.eq_ignore_ascii_case("dumb")))
+        .unwrap_or(false)
+}
+
+pub(crate) fn supports_progress() -> bool {
+    if !std::io::stderr().is_terminal() {
+        return false;
+    }
+    if term_is_dumb() {
+        return false;
+    }
+    if !std::io::stdin().is_terminal() {
+        return false;
+    }
+    true
+}
 
 pub(crate) static CONCURRENCY: LazyLock<usize> = LazyLock::new(|| {
     if EnvVars::is_set(EnvVars::PREK_NO_CONCURRENCY) {


yihong0618:
This is kind of the same problem of #1438 you can check my comments

crates/prek/src/printer.rs
```diff
@@ -43,7 +43,13 @@ impl Printer {
         match self {
             Self::Silent => ProgressDrawTarget::hidden(),
             Self::Quiet => ProgressDrawTarget::hidden(),
-            Self::Default => ProgressDrawTarget::stderr(),
+            Self::Default => {
+                if *crate::run::USE_COLOR {
```
j178: Does this mean that using `--color=never` will also turn off the progress bar?

yihong0618: yes that is what i am thinking color is not enough here.

smiling25261: The progress bar is not tied to color output , the solution is to explicitly disable progress reporting (for-example by using a dedicated "no-progress option" or conditional logic) instead of relying on  `--color=never`.

j178: There’s a global `--no-progress` option. What you’re saying doesn’t really seem relevant here, I think.

smiling25261: yeahh ,I was just giving an example to commit my words that `--color=never` doesn't really relying here.

shaanmajid: yeah this was a mistake; added a new fn for this

shaanmajid:
Pushed a fix and also added a stdin check which would cover scenarios like #1438

However, after investigating more, I think gating on stdin being a terminal is pretty non-standard for CLIs; uv or git don't do anything similar it looks like. #1438 looks mostly like user misconfiguration to me, and I don't feel strongly that we should handle those cases

I made an upstream fix which was merged in console-rs/indicatif#764, and it looks like they're gonna release that sometime within the next few days.

I'm leaning towards just closing this PR and wait for that release @j178 do you agree? If so you can go ahead and close :)

j178:
Yeah, I think we should not go in this way. Thanks for your work @shaanmajid!

PR #1562, opened 2026-02-04
test: skip symlink tests if create symlink failed

pcg0610:
- Added is_elevated crate to [target.'cfg(windows)'.dependencies]
- Required to check for admin privileges before running symlink tests

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -108,7 +108,6 @@ assert_cmd = { version = "2.0.16", features = ["color"] }
 assert_fs = { version = "1.1.2" }
 insta = { version = "1.40.0", features = ["filters"] }
 insta-cmd = { version = "0.6.0" }
-is_elevated = { version = "0.1" }
 markdown = { version = "1.0.0" }
 predicates = { version = "3.1.2" }
 pretty_assertions = { version = "1.4.1" }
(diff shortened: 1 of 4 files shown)


crates/prek/Cargo.toml
```diff
@@ -104,6 +103,9 @@ prek-pty = { workspace = true }
 libc = { workspace = true }
 pprof = { workspace = true, optional = true }
 
+[target.'cfg(windows)'.dependencies]
+is_elevated = { workspace = true }
```
j178: Should this go under `dev-dependecies`?

Cargo.toml
```diff
@@ -103,6 +102,7 @@ tracing-subscriber = { version = "0.3.20", features = ["env-filter"] }
 unicode-width = { version = "0.2.0", default-features = false }
 walkdir = { version = "2.5.0" }
 which = { version = "8.0.0" }
+is_elevated = { version = "0.1" }
```
j178: We should keep the dependencies in alphabetical order.

crates/prek/Cargo.toml
pcg0610: Done. Moved to dev-dependencies since it's only required for testing.

Cargo.toml
pcg0610: Sorry, I missed the alphabetical order. It's now fixed in the latest commit.

PR #1576, opened 2026-02-05
Update quick start to use `prek.toml`

j178:


YazdanRa:
Will there be a migration command in prek?
I would be more than happy to help add a migration command, if that's in the same direction as the vision.

j178:
It’ll be out by the end of the day :)

PR #1578, opened 2026-02-05
Include `prek.toml` in run hint for config filename

shaanmajid:


crates/prek/src/cli/run/run.rs
```diff
@@ -109,8 +109,9 @@ pub(crate) async fn run(
         if selectors.has_project_selectors() {
             writeln!(
                 printer.stderr(),
-                "\n{} If you just added new `{}`, try rerun your command with the `{}` flag to rescan the workspace.",
+                "\n{} If you just added a new `{}` or `{}`, try rerunning your command with the `{}` flag to rescan the workspace.",
                 "hint:".bold().yellow(),
+                "prek.toml".cyan(),
                 ".pre-commit-config.yaml".cyan(),
```
j178: Since you're here, it'd better to use constants from the prek-consts crate 😉

j178 (dismissed):

crates/prek/src/cli/run/run.rs
shaanmajid: Done!

j178 (approved):
Thanks!

PR #1580, opened 2026-02-06
Bump the supported uv version upper limit to `0.11.0`

j178:


shaanmajid:
Speaking of flaky tests... I'm pretty sure the Lua failures are representative of a legitimate race condition with the way Lua installs hooks. I'm taking a look into it

PR #1591, opened 2026-02-06
Resolve relative GIT_DIR and GIT_INDEX_FILE to absolute paths at startup

pb-dod:
## Problem

When `prek` is invoked with relative `GIT_DIR` or `GIT_INDEX_FILE` (e.g. `GIT_DIR=.git`), any internal CWD change such as `set_current_dir(workspace.root())` in `run.rs` or `--cd` causes git to resolve against the wrong directory. This is the same class of bug that stock pre-commit documents:

```python
# GIT_DIR: Causes git clone to clone wrong thing
# GIT_INDEX_FILE: Causes 'error invalid object ...' during commit
```

In practice, `git commit` passes absolute paths for these variables, so this isn't the root cause of the intermittent `fatal: unable to read <sha>` errors I've been seeing (see [#1591 comment](https://github.com/j178/prek/pull/1591#issuecomment-3862172032)). However, nothing guarantees these will always be absolute and other callers, git versions, or submodule configurations could pass relative paths.

## Fix

Resolve `GIT_DIR` and `GIT_INDEX_FILE` to absolute paths at startup in `main.rs`, before any CWD changes. This is placed immediately before the existing `GIT_WORK_TREE` stabilization logic.

## Tests

Two new integration tests in `hook_impl.rs`:
- `relative_git_dir_resolved_to_absolute` — verifies hooks see an absolute `GIT_DIR` when prek is invoked with `GIT_DIR=.git`
- `relative_git_index_file_resolved_to_absolute` — verifies the same for `GIT_INDEX_FILE=.git/index`

### DIFF
--- a/crates/prek/src/main.rs
+++ b/crates/prek/src/main.rs
@@ -200,7 +200,7 @@ async fn run(cli: Cli) -> Result<ExitStatus> {
     //      https://github.com/pre-commit/pre-commit/blob/8416413/pre_commit/git.py#L32-L35
     if let Some(git_dir) = EnvVars::var_os(EnvVars::GIT_DIR) {
         let git_dir_path = std::path::PathBuf::from(&git_dir);
-        if git_dir_path.is_relative() {
+        if git_dir_path.is_relative() && !git_dir_path.as_os_str().is_empty() {
             let abs_git_dir = std::path::absolute(&git_dir_path)
                 .context("Failed to resolve GIT_DIR to absolute path")?;
             debug!(
@@ -214,7 +214,7 @@ async fn run(cli: Cli) -> Result<ExitStatus> {
     }
     if let Some(index_file) = EnvVars::var_os("GIT_INDEX_FILE") {
         let index_path = std::path::PathBuf::from(&index_file);
-        if index_path.is_relative() {
+        if index_path.is_relative() && !index_path.as_os_str().is_empty() {
             let abs_index = std::path::absolute(&index_path)
                 .context("Failed to resolve GIT_INDEX_FILE to absolute path")?;
             debug!(


pb-dod:
Some more details about the issue I've been seeing:

The `fatal: unable to read` error hits multiple devs daily across every repo. It's always intermittent and always works on retry.

Our `.pre-commit-config.yaml` is pretty simple, just `pre-commit-hooks` (trailing-whitespace, end-of-file-fixer, check-merge-conflict) and a single hook from a private repo that runs `terraform fmt`. All hooks pass, then the post-hook `git diff` fails:

```
trim trailing whitespace.................................................Passed
fix end of files.........................................................Passed
check for merge conflicts................................................Passed
error: Command `git diff` exited with an error:
[status] exit status: 128
[stderr] fatal: unable to read 8b7b13987f665cf20b94ce324fed54091d7987e9
```

The SHA doesn't exist anywhere I can find. Not in the repo I'm committing to, not in `git verify-pack` on any pack file, not in any reflog or index, and it 404s on GitHub. It seems to only exist transiently during the failing run.

I initially thought this was related to partial/blobless clones (some of our tooling repos were cloned with `--filter=blob:limit=1m`) and lazy-fetch failures, but a coworker who only has full clones also hits this regularly so it doesn't seem specific to partial clones.

pb-dod:
Update: I applied the equivalent fix in our Python wrapper (resolving `GIT_DIR` and `GIT_INDEX_FILE` to absolute paths before spawning prek) and also disabled `gc.auto` via `GIT_CONFIG_COUNT` env vars. The error still occurred on a subsequent commit. However, I didn't verify whether `GIT_DIR` was actually set relative (or set at all) when it happened, so it's possible neither fix was exercised.

The failing SHA is a total ghost, not found in the committing repo, any prek cache repo, any local clone, any pack file, or prek's log. It only exists transiently during the failing run.

I still think this PR is a good defensive fix regardless since the CWD drift with a relative `GIT_DIR` is a real class of bug, but I can't confirm it addresses my specific intermittent issue yet. Happy to close if you'd prefer.

Issue #1583, opened 2026-02-06
Tracking issue for `prek.toml`


j178:
Implemetation:

- [x] https://github.com/j178/prek/commit/750cfbc7f3d72a7175487d232f68f02c0f0bdaf5
- [x] https://github.com/j178/prek/pull/1546
- [x] https://github.com/j178/prek/pull/1547
- [x] https://github.com/j178/prek/pull/1549
- [x] https://github.com/j178/prek/pull/1560 
- [x] https://github.com/j178/prek/pull/1576
- [x] https://github.com/j178/prek/pull/1578
- [x] https://github.com/j178/prek/pull/1584

Ecosystem upgrades:

- [x] Submit schema for `prek.toml` to SchemaStore: https://github.com/SchemaStore/schemastore/pull/5345
- [x] Add `prek.toml` to `prek-action` cache-key: https://github.com/j178/prek-action/pull/65
- [ ] Release new version of `prek`
- [ ] Release new version of `prek-action`
- [ ] Replace `.pre-commit-config.yaml` with `prek.toml` in `prek` and `prek-action` after release: https://github.com/j178/prek-action/pull/67
- [ ] Renovate support for `prek.toml`: https://github.com/renovatebot/renovate/discussions/40967

Related:
- [x] The original request for `prek.yaml`: https://github.com/j178/prek/issues/593
- [ ] Discussion about support configuration in `pyproject.toml` and `Cargo.toml`: https://github.com/j178/prek/issues/785

PR #1592, opened 2026-02-06
chore: replace .pre-commit-config.yaml with prek.toml

shaanmajid:
Ref #1583

Notes:
- Removed pre-commit from Renovate config; see https://github.com/renovatebot/renovate/discussions/40967 for tracking prek support
- Excluded `prek.toml` from Taplo, which does not yet suppport TOML 1.1
- Bumped prek-action to v1.1.1 ahead of Renovate schedule to pick up j178/prek-action#65


j178:
Let’s wait a bit and see how https://github.com/renovatebot/renovate/discussions/40967 goes before we end up losing Renovate support.

PR #1593, opened 2026-02-06
Make yaml-to-toml CONFIG argument optional

shaanmajid:
Now attempts to default to `.pre-commit-config.yaml` and `.pre-commit-config.yml` in PWD before erroring.

Feel free to close if you disagree, just ran into this when spinning up #1592, and figured it's worth fixing :)


j178 (approved):
Thanks!

PR #1601, opened 2026-02-07
Add `repo: self` support for consuming own hooks 

### LINKED ISSUE #1598: Add a `repo: self` option to allow easy consumption of hooks defined inside .pre-commit-hooks.yaml
By adding an option like `repo: self` or `repo: .`, that would not use `rev`, we could simplify self-testing of hook defined in current repository. This would help testing for hook publishers.

Current we are using other more convoluted methods for tests hooks defined locally while we are unable to use our own hooks inside the same repository that exposes them.

An alternative to this would be extend the meaning of the current `repo: local` to allow referencing hooks defined inside `pre-commit-hooks.yaml` without having to redefine them there (basically being able to mention just their id).

I hope this describes well enough the use case. I currently maintain a collection of hook at https://github.com/ansible/actions/blob/main/.pre-commit-hooks.yaml which rely on uv an npm native version management for reasons explained in #1462

shaanmajid:
Closes #1598

### DIFF
--- a/crates/prek/src/cli/run/run.rs
+++ b/crates/prek/src/cli/run/run.rs
@@ -1,8 +1,8 @@
 use std::fmt::Write as _;
 use std::io::Write as _;
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 use std::rc::Rc;
-use std::sync::{Arc, LazyLock};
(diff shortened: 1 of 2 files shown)


j178:
Let’s hold off on this for now. I think it adds unnecessary complexity for features that hardly get used.

PR #1610, opened 2026-02-09
Try to fix the flaky `color` test

j178:


shaanmajid:
I feel like it's been happening more lately? Would you agree?

j178:
Yeah, might be related to this part of code: https://github.com/j178/prek/blob/9a7d2e56afc20b0afd258de4bc10eb894c889b91/crates/prek/src/process.rs#L246-L256

PR #1614, opened 2026-02-09
Document `asdf` installation method

AlliotTech:
feat: add asdf install method  

asdf plugin repo: https://github.com/AlliotTech/asdf-prek  
related pr: https://github.com/asdf-vm/asdf-plugins/pull/1145


shaanmajid:
Thanks for the PR and the plugin setup, @AlliotTech!

Few notes:
1. Switched to draft status while waiting for upstream
2. I'm a bit concerned about the possibility of a supply-chain attack with keeping the plugin in a third-party repo.

AlliotTech:
Thanks for the review and for pointing out the concern about supply‑chain risk.

 The asdf‑prek plugin repository is a standard asdf plugin, maintained publicly on GitHub so that users can install via the usual asdf plugin add … steps.

Regarding the supply‑chain concern:
### Authenticity/visibility
The asdf‑prek repo is fully public — all code is visible and reviewable before anyone pulls it. There is no closed binary or opaque build step — users can inspect it and verify the install script before using it.

### Minimal scope
The plugin is essentially a wrapper to fetch releases of prek itself, not arbitrary third‑party code. All it does is support the official release artifacts and the documented install steps.

### Upstream/ownership options
If preferred, we can move the plugin into an organization‑controlled namespace (e.g., under j178 or the official prek org) so that the hosting is aligned with upstream and reduces trust concerns.
Security practices

We can additionally add a note in the docs recommending users verify the plugin repository’s commit history, its SHA sums, and point them to the official upstream releases — the same best practices used for any install helper script.

If you have specific requirements (e.g., signing the plugin, aligning with a canonical repo, or stricter vetting steps), I’m happy to address those. 

Would you prefer the plugin repository to be relocated under the j178/prek namespace before merging?

j178:
Thanks for the PR and for putting the asdf plugin together.

I’m going to close this for now. The upstream `asdf-plugins` PR does not seem to be moving, and that repo does not currently look like a reliable place for new plugin entries to be maintained.

Since this would document an installation method for `prek`, I’d prefer not to point users at an integration path unless it has a clear, maintained upstream story. Feel free to reopen this if the asdf plugin is accepted upstream or there is another stable maintenance path for it.

PR #1616, opened 2026-02-09
Add progress bar to `cache clean` and show removal summary

shaanmajid:
## Summary

- Print `Cleaning <path>` before deletion starts for immediate feedback
- Print `Removed N files (X.X MiB)` after completion
- Extract shared `DirStats` struct from `cache_size` for reuse
- Fix `human_readable_bytes(0)` returning NaN
- Silence spurious `fix_permissions` error when Go cache doesn't exist
- Standardize output formatting across cache commands (consistent spacing, styling, remove backticks)


j178:
Thank you!

PR #1636, opened 2026-02-14
Add `homeassistant-powercalc` to the list of users of prek

bramstroker:
Thanks for creating the tooling it's awesome.
Implemented in my project Powercalc, which is one of the most popular custom integration for Home Assistant.
Would be nice to list it in the reference list.


j178:
Thanks!

PR #1660, opened 2026-02-17
Sync identify file-type mappings with pre-commit identify

dbast:
Sync prek's file-type detection with identify by adding missing extension and filename mappings used by pre-commit. This is primarily additive, with a small set of compatibility tag expansions on existing mappings (for example msbuild and meson-options) so matching behavior is consistent; tests are extended to cover the new cases, including Tiltfile.

Well, I only need "Tiltfile" support for a local hook, but took the chance for a full sync. Happy to strip this down if this is too much. Thanks!


dbast:
the lua test failure seems unrelated ... maybe a re-run gets it green (its green in my repo fork [see](https://github.com/dbast/prek/actions/runs/22098681817/job/63862362492))

j178:
Thanks!

PR #1674, opened 2026-02-19
Install and compile gems in parallel

is-alnilam:
Currently the gem support code runs a single `gem install` to install the gems in the hook. Dependencies are resolved by the `gem` command, which installs each dependency in sequence. Any gems with native code are compiled, with the compilation stage of each gem using only a single compile task.

With this change, gems are installed in parallel (both top-level and dependencies), and any gems with native code allow these gems to use multiple compiler tasks (so using multiple cores where available). (Note that running multiple compilers will be the default `gem` behaviour in Ruby 4, so this ought to be safe).

The benefits of this will vary according to computer and hook, but an approximately 3x speed increase has been observed with the rubocop hook (from 38s down to 12s).

Additional speedup for https://github.com/j178/prek/issues/43


j178 (approved):
Thanks!

PR #1677, opened 2026-02-20
Jj workspace support

martijnberger:
## Motivation

I'm a happy jj (Jujutsu) user and I love prek. I just want the two to work together. Right now, prek fails in jj workspaces because:

1. **Secondary workspaces** (created with `jj workspace add`) have no `.git` directory, only `.jj/`. Every git command fails with "fatal: not a git repository".
2. **Even in colocated workspaces**, `git diff --staged` returns empty because jj doesn't use git's staging area. This makes the default `prek run` mode (which checks staged files) non-functional.

This PR makes prek detect jj workspaces transparently and do the right thing, with no configuration required.

### New: `crates/prek/src/jj.rs`

A new module for all jj-related logic, mirroring `git.rs` patterns:

`IS_JJ_WORKSPACE` detects `.jj/` by walking up from CWD. `detect_jj_git_dir()` resolves the backing `.git` directory from `.jj/repo/store/git_target`, handling both primary (colocated) and secondary workspaces. `setup_git_env_for_jj()` sets `GIT_DIR` and `GIT_WORK_TREE` env vars early in startup so all git commands work transparently. `get_changed_files()` uses `jj diff --name-only` to collect changed files in the working copy.

### Modified files

**`main.rs`** calls `setup_git_env_for_jj()` before any `GIT_DIR` handling.

**`cli/run/run.rs`** disables stashing for jj workspaces (stashing, `check_configs_staged()`, `has_unmerged_paths()`, and `WorkTreeKeeper::clean()` are all git-index-specific).

**`cli/run/filter.rs`** uses `jj::get_changed_files()` instead of `git::get_staged_files()` when in a jj workspace.

**`git.rs`** short-circuits `is_in_merge_conflict()` for jj (jj handles conflicts differently).

### Documentation

**`docs/faq.md`** new FAQ entry: "Does prek work with Jujutsu (jj)?"

**`docs/quickstart.md`** mentions jj working copy alongside git staging area.

**`docs/workspace.md`** updated repository boundary description to include `.jj`.

## Test plan

- [x] 7 new unit tests covering `find_jj_dir`, colocated workspace resolution, and secondary workspace resolution
- [x] All 288 existing unit tests pass
- [x] Zero clippy warnings on changed files (including pedantic + nursery lints)
- [x] Manual test in a jj colocated workspace with `prek run`
- [ ] Manual test in a jj secondary workspace with `prek run --all-files`
- [ ] Verify existing git behavior is unaffected (`IS_JJ_WORKSPACE = false` when no `.jj/` exists)


crates/prek/src/cli/run/filter.rs
```diff
@@ -387,8 +387,17 @@ async fn collect_files_from_args(
         return Ok(files);
     }
 
-    let files = git::get_staged_files(workspace_root).await?;
-    debug!("Staged files: {}", files.len());
+    let files = if *crate::jj::IS_JJ_WORKSPACE {
```
streadway: Rest of the project does not reference crates this way.

martijnberger: as in the crate:: prefix ?

crates/prek/src/cli/run/filter.rs
streadway, line 382: `jj file list` is the jj equiv.

martijnberger: Yeah I think I should take another pass and push this deeper into prek rather then patch the surface

j178:
Thanks! I definitely want to support `jj` in prek. But it needs some careful design and thought—not just patching things into the git code path.

Let’s mark this as a draft for now. Once I have a better understanding of `jj`, I’ll be in a much better position to review and maintain this code properly.

martijnberger:
> Thanks! I definitely want to support `jj` in prek. But it needs some careful design and thought—not just patching things into the git code path.

Agreed, ill take a proper pass at this tomorrow and do it by hand in the artisanal way

crates/prek/src/repo.rs
```diff
@@ -0,0 +1,317 @@
...
+    LazyLock::new(RepoContext::detect_current);
+
+/// Access the cached repository context with a normal `Result` API.
+///
+/// `LazyLock<Result<_>>` is convenient for one-time detection, but most callers
+/// want error propagation instead of reasoning about the lazy container directly.
+fn current() -> Result<&'static RepoContext> {
+    match REPO_CONTEXT.as_ref() {
+        Ok(repo) => Ok(repo),
+        Err(err) => Err(anyhow::anyhow!("{err}")),
```
nikosavola, line 97: It may be more more useful to also print a cause, see https://docs.rs/anyhow/latest/anyhow/struct.Error.html#display-representations

```suggestion
        Err(err) => Err(anyhow::anyhow!("{err:#}")),
```

Issue #1685, opened 2026-02-22
Proposal: Support python -m prek execution


uplsh580:
## Summary
prek can currently only be run via the prek command and does not work with `python -m prek`. Please add support for running prek as a Python module.

## Motivation
- Consistency: Many Python CLI tools (e.g. black, ruff, uv) support `python -m <package>`.
- Virtual environments: `python -m prek` makes it clear which Python interpreter (and thus which venv) is used.
- Scripts and CI: Using `python -m prek` makes the execution environment explicit.
- IDE and tool integration: Some tools only support the `python -m` form.

(I am willing to submit a PR)

CarliJoy:
> IDE and tool integration: Some tools only support the python -m form.

I am not aware of any. Can you name them?

> Virtual environments: python -m prek makes it clear which Python interpreter (and thus which venv) is used.

The whole idea is that prek creates isolated envs for each tool. So this is not really a use case.
Same is true for the script/ci argument.

uplsh580:
@CarliJoy 

Actually, This comment motivated me to open this PR.
https://github.com/apache/airflow/pull/62277#issuecomment-3939762269

CarliJoy:
Ok. That makes things clearer.

The use case is: Run prek installed in a python virtual env ensuring the prek version installed there is used.
Trying to simply use the bin PATH is more complicated if multiple OS need to be supported (i.e. windows .exe).

Still it took the liberty to suggest an alternative solution in the MR.

potiuk:
I am quite supportive of this. 

Let me add one more support the use-case: This came from a PR in Airflow's breeze dev env - where we make sure we have the **right** version of `prek` installed in `breeze` environment and not `prek` installed globally for the user. Which is actually important and covers one important use-case (not obvious) - currently `prek` supports `min_prek_version` (I suggested it in https://github.com/j178/prek/issues/421) - and that makes it possible for us in Airflow to start depending on new features of `prek` only available starting from version x.y.z.  But the way how min_prek_version works is that it will **fail** if minimum version is not available on path (which is expected if you run prek manually). And this adds friction.

In case of airflow `breeze` manages it's own virtualenv automatically, upgrading dependencies as needed, and if someone adds >=X.Y.Z for prek there, it will automatically reinstall itself (we expect breeze to be installed with `uv tool`). Which means that you will always get good prek version installed when you run prek hook via breeze. But prek available in PATH will be different - as prek is also installed manually for pre-commit integration for example. 

[...]

PR #1686, opened 2026-02-22
feat: add support for `python -m prek`

uplsh580:
Related Issue: https://github.com/j178/prek/issues/1685

## Summary
prek can now be run with python -m prek in addition to the prek command, matching common Python CLI tools like `black`, `ruff`, and `uv`.

## Motivation
- Consistency: Many Python tools support python -m <package>
- Virtual environments: Makes it clear which Python interpreter (and venv) is used
- Scripts & CI: Explicit execution environment
- Tool integration: Some tools only support the `python -m` form

## Changes
- Add `python/prek/` package with `__init__.py` and `__main__.py`
- Set python-source = "python" in `pyproject.toml` so maturin includes the Python package
- `__main__.`py invokes the prek binary next to the Python executable (e.g. `.venv/bin/prek with .venv/bin/python`), falling back to PATH if not found

### DIFF
--- a/python/prek/__main__.py
+++ b/python/prek/__main__.py
@@ -3,26 +3,24 @@ import shutil
 import subprocess
 import sys
 
+from ._find_prek import find_prek_bin
 
-def main() -> None:
-    # With `python -m prek`, runpy sets sys.argv[0] to the module path and argv[1:] to the args
-    prek_args = sys.argv[1:]
+def _run() -> None:
+    prek = find_prek_bin()
 
-    # Prefer prek next to the Python executable so that `python -m prek` uses the
-    # same prek as that Python environment (e.g. .venv/bin/prek with .venv/bin/python)
-    python_dir = os.path.dirname(os.path.abspath(sys.executable))
-    prek_bin = os.path.join(python_dir, "prek")
     if sys.platform == "win32":
-        prek_bin += ".exe"
+        import subprocess
 
-    if not os.path.isfile(prek_bin):
-        prek_bin = shutil.which("prek")
-    if prek_bin is None:
-        print("prek: command not found", file=sys.stderr)
-        sys.exit(127)
+        # Avoid emitting a traceback on interrupt
+        try:
+            completed_process = subprocess.run([prek, *sys.argv[1:]], env=env)
+        except KeyboardInterrupt:
+            sys.exit(2)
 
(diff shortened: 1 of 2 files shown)


uplsh580:
# Validation
image

uplsh580:
Actually, This comment motivated me to open this PR.
https://github.com/apache/airflow/pull/62277#issuecomment-3939762269

uplsh580:
https://github.com/j178/prek/actions/runs/22267961503/job/64417541100?pr=1686

I think the `prek::run color` failure does not appear to be related to this PR. The changes here only add Python package support for `python -m prek` and do not modify any Rust code or the prek run logic. The color test checks ANSI color output when running the prek binary directly, and the snapshot mismatch is likely due to CI environment differences (e.g., terminal emulation or how `cargo-llvm-cov` captures output) rather than these changes.

python/prek/__main__.py
```diff
@@ -0,0 +1,28 @@
...
+    if sys.platform == "win32":
+        prek_bin += ".exe"
+
+    if not os.path.isfile(prek_bin):
+        prek_bin = shutil.which("prek")
+    if prek_bin is None:
+        print("prek: command not found", file=sys.stderr)
+        sys.exit(127)
+
+    sys.exit(subprocess.run([prek_bin] + prek_args).returncode)
```
yihong0618: I think we can consider use `ruff` way
the code is simple

https://github.com/astral-sh/ruff/blob/main/python/ruff/__main__.py

uplsh580: @yihong0618 Thanks for your comment.
https://github.com/astral-sh/ruff/blob/66defe95828a65ea41d6c565305349171f536451/python/ruff/_find_ruff.py#L11-L55
This code needs to be implemented further, but for now, the code in the current PR seems simpler.

j178:
Thanks!

python/prek/__main__.py
```diff
@@ -0,0 +1,26 @@
+import os
+import shutil
```
CarrotManMatt, line 2:
```suggestion
```
Import is not used.

python/prek/__main__.py
```diff
@@ -0,0 +1,26 @@
...
+import subprocess
```
CarrotManMatt, line 3:
```suggestion
```
Import is only used if `sys.platform == "win32"` and is correctly conditionally imported on line 9.

PR #1692, opened 2026-02-25
Install hooks to core.hooksPath instead of refusing

### LINKED ISSUE #23: Correctly support `core.hooksPath`

domenkozar:
## Summary

- When `core.hooksPath` is set, install hooks to that directory instead of bailing with "Cowardly refusing"
- This makes `prek install` idempotent and ensures hooks end up where git will actually look for them
- If the configured path is not writable, bail with a clear error message

## Motivation

Tools like [devenv](https://devenv.sh) set `core.hooksPath` to manage git hooks. When prek is used as one of those hooks, running `prek install` a second time fails because it sees `core.hooksPath` is already set. The current behavior also means hooks installed to `.git/hooks` would never run when `core.hooksPath` points elsewhere.

Instead of refusing, prek should install to the configured path (where git will actually look for hooks). This resolves #23.

## Test plan

- [ ] `prek install` with `core.hooksPath` unset works as before
- [ ] `prek install` with `core.hooksPath` set to a writable directory installs hooks there
- [ ] Running `prek install` twice with `core.hooksPath` set succeeds (idempotent)
- [ ] `prek install` with `core.hooksPath` set to a read only directory gives a clear error

🤖 Generated with [Claude Code](https://claude.com/claude-code)


domenkozar:
@j178 we switched over to prek, but this one is annoying because prek wants to force not setting `core.hooksPath` but some other tools want to set it.

domenkozar:
@j178 curious what you think about this solution :)

j178:
Thanks! I haven’t had much time for open source lately, but I’ll get back to you soon.

janlarres:
I don't think silently writing to a global hooks path is a good idea. This will obviously affect all repositories which may not be desired, especially since prek will return an error if a repository doesn't have a config file. It may also overwrite a pre-existing global hook.

For example, in my case I have [a hook script](https://github.com/janlarres/etc/blob/main/dotfiles/git/hooks/stub) that you symlink to the hook names like `pre-commit` and which then runs all files in the respective `*.d` directory like `pre-commit.d`, in both the global hooks dir and the local one. This means that prek works fine for my setup as long as I unset `core.hooksPath`, run `prek install`, and set the path again, but this change would silently break all of my other pre-commit hooks by overwriting the symlink.

Personally my ideal solution would be if I could set a global config option to run `prek install` despite `core.hooksPath` being set because my setup is designed to deal with that, but this may not work for everyone.

domenkozar:
> I don't think silently writing to a global hooks path is a good idea. This will obviously affect all repositories which may not be desired, especially since prek will return an error if a repository doesn't have a config file. It may also overwrite a pre-existing global hook.
> 
> For example, in my case I have [a hook script](https://github.com/janlarres/etc/blob/main/dotfiles/git/hooks/stub) that you symlink to the hook names like `pre-commit` and which then runs all files in the respective `*.d` directory like `pre-commit.d`, in both the global hooks dir and the local one. This means that prek works fine for my setup as long as I unset `core.hooksPath`, run `prek install`, and set the path again, but this change would silently break all of my other pre-commit hooks by overwriting the symlink.
> 
> Personally my ideal solution would be if I could set a global config option to run `prek install` despite `core.hooksPath` being set because my setup is designed to deal with that, but this may not work for everyone.

I understand why your workflow works the way it does, but it feels backwards that git providers a way to tell where to install git hooks and then prek just bails out when that option is set.

I'd suggest that by default prek supports the option, optionally warns about where it's installing the hook and allows a knob to disable respecting the option for such cases. Would that work for you?

domenkozar:
The issue I'm having is that https://github.com/cachix/git-hooks.nix sets `core.hooksPath` to `.git/hooks` so running prek install bails out where `pre-commit` respects that.

j178:
> prek install sets core.hooksPath to .git/hooks

No, I don’t think prek changes any Git configuration. What makes you think that?

domenkozar:
> > prek install sets core.hooksPath to .git/hooks
> 
> No, I don’t think prek changes any Git configuration. What makes you think that?

My mistake, I've changed the description of what happens :)

j178:
> where pre-commit respects that

This doesn’t seem right either. Since prek uses the same logic as pre-commit, it should bail out too. Could you take a look at that?

```console
❯ git init
❯ git config core.hooksPath hooks
❯ pre-commit install
[ERROR] Cowardly refusing to install hooks with `core.hooksPath` set.
hint: `git config --unset-all core.hooksPath`
```

j178:
@domenkozar Could you take a look at #1723? It adds a `--git-dir` flag to `prek install` so you can force the hook installation target directory.

Issue #1696, opened 2026-02-26
priority: parallel hooks will end-up triggering Too many open files (os error 24)


ssbarnea:
### Summary

I spotted this issue with multiple hooks from https://github.com/pre-commit/pre-commit-hooks.git 

Apparently commenting one or two of them or using different priorities is the only workaround I was able to find.

```
error: Failed to run hook `trailing-whitespace`
  caused by: Failed to setup subprocess for pty
  caused by: Too many open files (os error 24)
```

### Willing to submit a PR?

- [ ] Yes — I’m willing to open a PR to fix this.

### Platform

macos-latest

### Version

prek 0.3.3 (c741c8c27 2026-02-15)

### .pre-commit-config.yaml

```
hooks:
  - repo: https://github.com/pre-commit/pre-commit-hooks.git
    rev: v6.0.0
    hooks:
      - id: check-added-large-files
        priority: 0
      - id: check-merge-conflict
        priority: 0
      - id: check-symlinks
        priority: 0
      - id: debug-statements
        priority: 0
      - id: detect-private-key
        priority: 0
      - id: end-of-file-fixer
        priority: 0
      - id: trailing-whitespace
        priority: 0
```

### Log file

prek run -a

drichardson:
In concurrency mode, perhaps prek should take into account open file descriptors when deciding how concurrent it can be.

j178:
I think we could also look at adjusting the ulimit on startup, similar to what uv did here: https://github.com/astral-sh/uv/pull/17464

drichardson:
Yeah that would be ideal. It was the first approach I thought about taking, but the change was quite a bit bigger and a little tweaky, you need to know the max number of file descriptors that can be opened per concurrent process... and if someone makes a change to prek later that increases/decreases that max, you want to update it.

drichardson:
Thank you for taking the #1697 so quickly @j178. What's the cadence for releases for this project? One of our (Whatnot's) primary repos use this and I had to disable concurrency, to the dismay of my engineers, so I'm looking forward to using this mitigation to re-enable concurrency.

We've been loving prek as a replacement to pre-commit btw, well done and thanks.

drichardson:
@ssbarnea this change is released, as a mitigation (until auto ulimit detection comes if it does) you can set PREK_MAX_CONCURRENCY=X, where X is some number that makes the "Too many open files" problem not happen. For me, on a 16 core mac with my config file, it happens to be 14.

PR #1697, opened 2026-02-27
Add PREK_MAX_CONCURRENCY environment variable

drichardson:
## Summary

- Add `PREK_MAX_CONCURRENCY` environment variable to cap the maximum number of concurrent hooks
- Value is floored at 1; values above CPU count are allowed (useful for I/O-bound hooks)
- Invalid values show a user-visible warning and fall back to CPU count
- `PREK_NO_CONCURRENCY` takes precedence when both are set

## Motivation

When `ulimit -n` is low, concurrent hook execution can exhaust file descriptors. This provides an environment variable to limit concurrency.

For #1696

## Test plan

- [x] `mise run lint` — no warnings
- [x] `mise run test` — all tests pass
- [x] Unit tests for `resolve_concurrency`: valid value, above CPU count, zero floors to 1, invalid string, empty string, unset, no_concurrency, no_concurrency overrides max


j178:
Thanks!

Issue #1731, opened 2026-03-01
Race condition in Ruby parallel gem installation


is-alnilam:
### Summary

There seems to be a race condition in the parallel gem installation code, that's especially likely to be seen on Windows. Sorry, it didn't show up in the initial CI runs, but has cropped up a couple of times since...

Basically, two `gem` processes are trying to write to (different files in) the same folder at the same time. Then they try to read *all* the files in that folder. If one of the processes is still working on writing a file when the other process reads it, it gets corrupt data and errors.

This isn't a major issue on Linux/MacOS, mostly as, on these platforms, `gem` uses write-and-rename semantics to atomically update the file. It seems that Windows isn't so good at this.

Sadly, all of my stress testing of the change was on Linux, with only the CI runs to check the Windows behaviour.

One option is obviously to back out the parallel gem support, or to default it to run in serial mode rather than parallel. But I've also got a PR for a hardened approach which I'll raise in a moment.

### Willing to submit a PR?

- [x] Yes — I’m willing to open a PR to fix this.

### Platform

windows-latest CI runners

### Version

prek 0.3.4

### .pre-commit-config.yaml

```yaml
        repos:
          - repo: local
            hooks:
              - id: test-gem-require
                name: test-gem-require
                language: ruby
                entry: ruby test_script.rb
                language_version: system
                additional_dependencies: ["rspec"]
                pass_filenames: false
                always_run: true
              - id: test-gem-require-versioned
                name: test-gem-require-versioned
                language: ruby
                entry: ruby test_script.rb
                language_version: system
                additional_dependencies: ["rspec:3.12.0"]
                pass_filenames: false
                always_run: true
              - id: test-gem-require-missing
                name: test-gem-require-missing
                language: ruby
                entry: ruby test_script.rb
                language_version: system
                pass_filenames: false
                always_run: true
```

### Log file

```
        FAIL [  13.336s] ( 6/12) prek::languages ruby::additional_gem_dependencies
  stdout ───

    running 1 test
    ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ Snapshot Summary ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
    Snapshot: additional_gem_dependencies
    Source: D:\a\prek\prek:325
    ───────────────────────────────────────────────────────────────────────────────
    program: prek
    args:
      - run
      - "-v"
    env:
      GIT_CONFIG_COUNT: "1"
[...]

PR #1732, opened 2026-03-01
Allow parallel gem retry

is-alnilam:
Fix for #1731

Especially on Windows, installing Ruby gems in parallel can fail with a race condition as `gemspec` files are written simultaneously, and can be read in a part-written form. This change implements a retry/fallback strategy to handle this and still keep the majority of the performance gains of the parallel installation path.

1. Try installing in parallel as before.
2. If there's an error (of any kind), try each gem again after a short, random delay. The delay is there to try to introduce a bit of jitter, in case multiple gems all failed and they all then try to install at exactly the same time again.
3. If that fails too, automatically fall back to sequential install. This will re-use any gems that were already installed in parallel.

In the happy path case, this has no impact to runtime. This will almost always be the case on Unix-based systems.

The next path has a slight slow-down due to the retry and the pause, but it's still pretty quick.

And in the final case, hopefully a fair amount of the work has already been done.

The drawback of the above, of course, is that it's more complex than the basic parallel path, and even that's more complex than the basic sequential process.

The change also includes a tweak to the tests, where the 'waiting for lock' message occasionally appeared (which may coincide with occasions when the retry behaviour above was triggered). The warning is now filtered out and ignored by the tests.

As this is a race condition it's hard to guarantee that this fixes the problem, but I've run 15 cycles of the windows/ruby test in CI and have seen no failures.

### DIFF
--- a/crates/prek/tests/common/mod.rs
+++ b/crates/prek/tests/common/mod.rs
@@ -415,6 +415,8 @@ pub const INSTA_FILTERS: &[(&str, &str)] = &[
     ),
     // Time seconds
     (r"\b(\d+\.)?\d+(ms|s)\b", "[TIME]"),
+    // Strip non-deterministic lock contention warnings from parallel test execution
+    (r"(?m)^warning: Waiting to acquire lock.*\n", ""),
 ];
 
 #[allow(unused_macros)]


j178:
Thanks!

PR #1755, opened 2026-03-08
Respect `core.sharedRepository` for hook permissions

tennox:
I use git repos with multiple users, but encounter problems like:
```
--- devenv:git-hooks:install stderr:
0000.00: error: failed to open file `.git/hooks/pre-commit`: Permission denied (os error 13)
```

So I adapted this tool to respect the `core.sharedRepository` config.

When `core.sharedRepository` is set to `group` (or similar), hook files are now
created with `0o775` instead of `0o755`, allowing multiple users to update them.

### DIFF
--- a/crates/prek/src/cli/install.rs
+++ b/crates/prek/src/cli/install.rs
@@ -46,7 +46,9 @@ pub(crate) async fn install(
         );
     }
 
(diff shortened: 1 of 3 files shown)


tennox:
@codecov refresh

j178:
Thanks!

PR #1758, opened 2026-03-09
Improve benchmark script

j178:


kushudai:
Thank you and sorry about that HEAD_BINARY bug

j178:
@kushudai No worries, I think that is introduced by me :)

PR #1764, opened 2026-03-09
Update configuration.md with TOML 1.1 notes

pcastellazzi:
TOML 1.1 is not yet universally available. This hit me on Visual Studio Code using [Even Better TOML](https://marketplace.visualstudio.com/items?itemName=tamasfe.even-better-toml). While this issue does not affect `prek` directly, users may encounter errors on their IDE/editor and wonder what is going on.


j178:
Thanks!

PR #1812, opened 2026-03-16
Modify 'cuh' TagSet to include 'header'

HagegeR:
Updated the TagSet for 'cuh' to include 'header' in addition to 'text'.


j178:
Hi! This file is auto-generated from the Python [identify](https://github.com/pre-commit/identify) package, so it shouldn’t be edited manually. Could you please open a PR upstream instead?

HagegeR:
opened https://github.com/pre-commit/identify/pull/584
thanks for the input!

PR #1817, opened 2026-03-17
Add `--all` flag to `prek uninstall`

### LINKED ISSUE #1779: `prek uninstall` without `-t` should remove all prek-managed hooks
### Summary

`prek uninstall` without `-t` only removes the `pre-commit` hook, even if other prek-managed hooks (e.g. `pre-push`, `commit-msg`) exist in `.git/hooks/`. Users have to remember exactly which `-t` flags they used during install, or set `default_install_hook_types` in the config, to fully uninstall.

This is the same issue as [pre-commit/pre-commit#364](https://github.com/pre-commit/pre-commit/issues/364).

**Note**: This is specifically about the **uninstall** behavior, not about auto-detecting hook types during install (which was discussed in #1774 / #1224).

### Reproduction

```bash
prek install -t pre-push
prek uninstall
# .git/hooks/pre-push still exists and is managed by prek
```

### Expected behavior

`prek uninstall` without `-t` (and without `default_install_hook_types` in config) removes **all** prek-managed hooks found in the hooks directory.

prek already has `is_our_script()` to detect its own hooks, so scanning the hooks directory and removing only prek-managed ones is straightforward and safe.

### Current behavior

`get_hook_types()` falls back to `[pre-commit]`, so only the `pre-commit` hook is targeted for removal.

### Proposed fix

Add a separate resolution path for uninstall: when no `-t` is given and no `default_install_hook_types` is configured, iterate over all known `HookType` variants in the hooks directory and remove those where `is_our_script()` returns true.

This keeps the current behavior when `-t` or `default_install_hook_types` is explicitly set.

I have a draft implementation on my fork: [`Tiryoh/prek@fix/uninstall-remove-all-managed-hooks`](https://github.com/Tiryoh/prek/tree/fix/uninstall-remove-all-managed-hooks).
If this direction looks good, I'm happy to open a PR as-is.

### Willing to submit a PR?

- [x] Yes — I'm willing to open a PR to fix this.

### Platform

Linux

### Version

prek 0.3.5

Tiryoh:
## Summary

  Closes #1779

  - Add `--all` flag to `prek uninstall` that scans the hooks directory and removes every prek-managed hook (detected via `is_our_script()`), regardless of hook type
  - Without `--all`, behavior is unchanged: uses `-t` or `default_install_hook_types`, falls back to `pre-commit`
  - `--all` and `-t` are mutually exclusive (`conflicts_with`)
  - Suppress "not managed" / "does not exist" skip messages when `--all` is used, since iterating all hook types will naturally hit many non-existent ones

  ## Tests

  - `uninstall_all_managed_hooks`: `--all` removes all prek-managed hooks
  - `uninstall_all_no_hooks`: `--all` exits cleanly when no hooks are installed
  - Ran `mise run lint` and confirmed no errors

### DIFF
--- a/crates/prek/src/cli/install.rs
+++ b/crates/prek/src/cli/install.rs
@@ -5,6 +5,7 @@ use std::sync::Arc;
 
 use anyhow::{Context, Result};
 use bstr::ByteSlice;
+use clap::ValueEnum;
 use owo_colors::OwoColorize;
 use prek_consts::CONFIG_FILENAMES;
 use same_file::is_same_file;
@@ -12,8 +13,6 @@ use same_file::is_same_file;
 use crate::cli::reporter::{HookInitReporter, HookInstallReporter};
 use crate::cli::run;
 use crate::cli::run::{SelectorSource, Selectors};
-use clap::ValueEnum;
-
 use crate::cli::{ExitStatus, HookType};
 use crate::config::load_config;
 use crate::fs::{CWD, Simplified};
--- a/crates/prek/tests/install.rs
+++ b/crates/prek/tests/install.rs
@@ -765,7 +765,7 @@ fn uninstall() -> anyhow::Result<()> {
 
 /// `prek uninstall --all` should remove all prek-managed hooks.
 #[test]
-fn uninstall_all_managed_hooks() {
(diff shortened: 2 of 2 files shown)


j178:
Thank you!

Issue #1707, opened 2026-02-28
"Failed to create Python virtual environment" because of unknown missing file


l0b0:
### Summary

Just testing out prek with Nix, and it's working great locally. But [in CI I get the following error](https://gitlab.com/engmark/root/-/jobs/13305020918):

```
error: Failed to install hook `pretty-format-ini`
  caused by: Failed to create Python virtual environment
  caused by: Run command `create venv` failed
  caused by: No such file or directory (os error 2)
```

### Willing to submit a PR?

- [ ] Yes — I’m willing to open a PR to fix this.

### Platform

NixOS x86_64

### Version

prek 0.3.0

### .pre-commit-config.yaml

```yaml
---
# Configuration file for pre-commit (https://pre-commit.com/).
# Please run `pre-commit run --all-files` when adding or changing entries.

# Put formatters before linters to avoid spurious error messages
repos:
  - repo: https://github.com/jamescherti/pre-commit-elisp
    rev: 1415c3c1f0dbad79abb3ec86a6158681a1ab681d # frozen: v1.0.8
    hooks:
      - id: elisp-check-parens
      - id: elisp-byte-compile
      - id: elisp-indent

  - repo: https://github.com/macisamuele/language-formatters-pre-commit-hooks
    rev: 4380fbb73a154b5f5624794c1c78d9719ccc860f # frozen: v2.16.0
    hooks:
      - id: pretty-format-ini
        args: [--autofix]

  - repo: https://github.com/pre-commit/pre-commit-hooks
    rev: 3e8a8703264a2f4a69428a0aa4dcb512790b2c8c # frozen: v6.0.0
    hooks:
      - id: pretty-format-json
        args: [--autofix]
      - id: pretty-format-json
        args: [--autofix]
        files: ^includes/mcomix/.*\.conf$
        types: []

  - repo: https://gitlab.com/engmark/sort-hook
    rev: f7c2fa555b1c3a32132b9b1a8f2b852c919dd4e7 # frozen: v5
    hooks:
      - id: sort
        args: [--locale=en_NZ.UTF-8]
        files: ^(.*/\.gitignore|includes/(\.bash_history|darktable/(keyboardrc|shortcutsrc|shortcutsrc\.defaults)|repl-history))$
        stages: [pre-commit]
      - id: sort
        args:
          [
            "--sort-option=--field-separator=\t",
            --sort-option=--key=2,
            --sort-option=--key=1,
            --locale=en_NZ.UTF-8,
          ]
[...]

PR #1824, opened 2026-03-19
python: move uv tests into the test module

shaanmajid:
## Summary

- Validate the downloaded uv binary immediately after install.
  - Now, a non-functional binary (e.g. glibc/ELF interpreter mismatch on NixOS) fails early during hook installation, rather than on hook invocation.
  - Failure is attributed to the binary itself rather than surfacing as a confusing venv creation error.
- Wrap existing bare `#[test]` functions in `#[cfg(test)] mod tests` to match codebase conventions.

Ref #1707

### DIFF
--- a/crates/prek/src/languages/python/uv.rs
+++ b/crates/prek/src/languages/python/uv.rs
@@ -573,149 +573,154 @@ fn uv_source_from_env() -> Option<InstallSource> {
     }
 }
 
-#[test]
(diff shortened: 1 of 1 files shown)


j178:
Haha, you beat me by 5 minutes 😄 Here’s mine: https://github.com/j178/prek/pull/1825

j178:
Thanks! I repurposed this PR to only be about moving tests.

PR #1837, opened 2026-03-20
docs: fix incorrect type reference in ArchiveExtension::from_path

stringsbuilder:
Corrected the doc comment for `ArchiveExtension::from_path` which incorrectly  referenced `SourceDistExtension`. 

The method returns `ArchiveExtension`, not `SourceDistExtension`.


j178:
Thanks!

PR #1844, opened 2026-03-21
Add `format-ipy-cells` builtin hook

janosh:
## Summary

- Adds a Rust-native builtin hook `format-ipy-cells` that formats `# %%` cell delimiters in [VS Code interactive Python notebooks](https://code.visualstudio.com/docs/python/jupyter-support-py)
- Based on the Python [format-ipy-cells](https://github.com/janosh/format-ipy-cells) hook, reimplemented using a **structured line-based parser** (parse -> format -> serialize) instead of chained regex substitutions, making it more robust and easier to reason about
- Includes 21 tests covering parsing, individual formatting rules, merge behavior, idempotency, and async file I/O

### Formatting rules

- Normalize delimiter spacing: `#%%`, `#   %%` -> `# %%`
- Normalize comments: `# %%comment`, `# %%   comment` -> `# %% comment`
- Remove empty cells (no comment and no code)
- Merge bare delimiters separated by only whitespace into the previous cell
- Remove leading/trailing blank lines within each cell
- Ensure exactly two blank lines before each cell delimiter
- One blank line between a triple-quoted string and the following cell (ruff compatibility)
- Remove trailing empty cell at end of file
- Strip trailing whitespace from each line

### Usage

```yaml
repos:
  - repo: builtin
    hooks:
      - id: format-ipy-cells
```

No Python environment needed -- runs as native Rust code inside prek.

## Test plan

- [x] All 21 unit tests pass (`cargo test -p prek format_ipy_cells`)
- [x] Full fixture test validates end-to-end formatting against known good output
- [x] Idempotency test confirms formatting a clean file produces identical output
- [x] Mutation testing verified all 6 key code paths are caught by tests
- [x] Existing prek test suite passes (356 unit tests + all builtin hook integration tests)
- [ ] Manual testing with real-world interactive Python notebooks

### DIFF
--- a/prek.schema.json
+++ b/prek.schema.json
@@ -890,6 +890,7 @@
         "detect-private-key",
         "end-of-file-fixer",
         "fix-byte-order-marker",
+        "format-ipy-cells",
         "mixed-line-ending",
         "no-commit-to-branch",
         "trailing-whitespace"


j178:
Thanks for putting this together, the implementation and test coverage look solid. Before merging, I’d like to understand the expected adoption a bit better.

Adding a new builtin hook means committing to maintaining it long term in prek, so I want to make sure this is addressing a broadly used workflow rather than a niche case.

Do you have a sense of how widely this hook is used, or any signals that it has meaningful adoption beyond a smaller set of users?

janosh:
there's next to no existing user base for this hook. this cell delimeter syntax is quite widely used though thanks to long-standing VS Code support https://code.visualstudio.com/docs/python/jupyter-support-py#_jupyter-code-cells

mostly used by people who do Python data analysis but want something lean to track in version control (i.e. not `.ipynb`) and easy to run in CI to ensure demo notebooks don't break over time as the code being demoed evolves ([example](https://github.com/TorchSim/torch-sim/blob/1bcd95d13/examples/tutorials/high_level_tutorial.py)). these are frequent use cases so i think this hook could garner a small to medium sized user base if offered in `prek`.

re maintenance effort, my experience with it has been that it's very low (see commit history in https://github.com/janosh/format-ipy-cells). that said, i understand if this is too niche. feel free to close

janosh:
@j178 let me know if you need more info for a decision. i certainly understand if you don't think this fits in `prek`

j178:
Thanks for putting this together. I appreciate the work here, but I’m going to pass on merging this PR.

My main concern is that I don’t think this hook is broad enough to be a good fit as a built-in hook in prek. It feels a bit too specific for that.
Thanks again for taking the time to contribute.

PR #1863, opened 2026-03-25
Add `--exclude` option and `pin` config field for auto-update

thatmlopsguy:
This PR adds two complementary ways to skip repositories during `prek auto-update`. Inspired by similar project [pre-commit-update](https://gitlab.com/vojko.pribudic.foss/pre-commit-update).

- -e, --exclude — Exclude repos by substring match at runtime
- pin config field — Permanently pin a repo to skip it during auto-update

Both are useful when you need to stay on a specific version of a hook repository — for example, while upgrading your codebase to be compatible with a newer release.

```
# Exclude a single repo
prek auto-update --exclude ruff

# Exclude multiple repos
prek auto-update --exclude ruff --exclude isort

# Exclude all repos (useful for dry-testing other flags)
prek auto-update --exclude '*'
```

Using --exclude '*' also prevents combining with `--bleeding-edge`, since updating nothing to bleeding edge is contradictory.

A per-repo boolean field that causes auto-update to silently skip the repository:

```toml
[[repos]]
repo = "https://github.com/psf/black"
rev = "24.1.0"
pin = true
hooks = [{ id = "black" }]
```

```yaml
repos:
  - repo: https://github.com/psf/black
    rev: v24.1.0
    pin: true
    hooks:
      - id: black
```


j178:
I think we should discuss these features in the issue first. Also, I’d prefer to split `--exclude` and `pin` into separate PRs.

j178:
Thanks for working on this. It’s partially superseded by #1983, and if we still want the `pin` feature, we can open a new issue to discuss it.

thatmlopsguy:
Sorry, but I have to find sometime to work on this. Thanks.

PR #1874, opened 2026-03-31
Remove unnecessary rust version in CONTRIBUTING.md

geyerandreas:
The current Rust version listed in the `CONTRIBUTING.md` file is incorrect.
According to `rust-toolchain.md` the version is pinned to `1.94`
- #1751 

Removing the listed version in `CONTRIBUTING.md` completely prevents using the wrong rust version.
I guess one can expect a developer to be able to look up the pinned version in `rust-toolchain.md` directly.


j178:
Thanks!

Issue #1868, opened 2026-03-27
Failed hooks can be reported as "Passed" when in a `priority` parallelism


nathanjmcdougall:
### Summary

I'm finding that if a hook fails, then it will be reported as "Passed" when being executed in parallel with another hook at the same priority level. Perhaps this is intended behaviour but I think it ought to at least be documented better.

To reproduce: consider an otherwise empty git repo with file `violates.py` containing bad formatting:

```Python
x = (
    2 + 
     11
)
```

Then with this `.pre-commit-config.yaml` file:

```YAML
minimum_prek_version: 0.2.23
repos:
  - repo: https://github.com/astral-sh/ruff-pre-commit
    rev: v0.14.3
    hooks:
      - id: ruff-check
        priority: 0
      - id: ruff-format
        priority: 0
```

I get this output when running the hooks:

```
uvx prek@latest -a                                                                       
Files were modified by following hooks...................................Failed
  ┌ ruff check...........................................................Passed                                      
  └ ruff format..........................................................Passed
```

I would have expected `ruff format` to be labelled `Failed` here, not `Passed`, and to present the user with the stderr from `ruff format`.

If I change `ruff-format` to `priority: 1` or remove the `priority` config entirely, I get:

```
uvx prek@latest                                                                          
ruff check...............................................................Passed
ruff format..............................................................Failed                                      
- hook id: ruff-format                                                                                               
- files were modified by this hook

  1 file reformatted
```

Which is the expected behaviour.

### Willing to submit a PR?

- [x] Yes — I’m willing to open a PR to fix this.

### Platform

Windows 11 x86_64

### Version

prek 0.3.8 (bb412c080 2026-03-23)

### .pre-commit-config.yaml

```yaml
minimum_prek_version: 0.2.23
repos:
  - repo: https://github.com/astral-sh/ruff-pre-commit
    rev: v0.14.3
    hooks:
      - id: ruff-check
        priority: 0
      - id: ruff-format
        priority: 0
```

### Log file

```
2026-03-27T00:31:49.565628Z DEBUG prek: 0.3.8 (bb412c080 2026-03-23)
2026-03-27T00:31:49.565739Z DEBUG Args: ["C:\\Users\\namc\\AppData\\Local\\uv\\cache\\archive-v0\\TdnOTdnzPiu-p5gN1P3xd\\Scripts\\prek.exe"]
2026-03-27T00:31:49.639005Z TRACE get_root: close time.busy=73.2ms time.idle=9.10µs
[...]

PR #1879, opened 2026-04-01
Update docs for case of hooks modifying files with a non-zero exit code

nathanjmcdougall:
See https://github.com/j178/prek/issues/1868 for context.


j178:
Thanks!

PR #1919, opened 2026-04-09
Handle impostor commits in auto-update

### LINKED ISSUE #1864: feat: add utility to check hashes and frozen: tags match
If you use `--freeze`, which adds a little security (see recent litellm disaster), it also introduces a problem keeping them in sync. For example, say I have this:

```yaml
 - repo: https://github.com/pre-commit/pre-commit-hooks
   rev: 3e8a8703264a2f4a69428a0aa4dcb512790b2c8c  # frozen: v6.0.0
```

In GitHub, if you make a fork, your comments are also accessible as an SHA on the upstream. So, for example, I could fork the above repo, get a new hash, and then open a seemingly innocuous PR that puts a fake take on a `rev` from downstream. ~~I _think_ pre-commit and prek are both immune to this problem since they actually fetch the repo~~ (GitHub Actions is not! And nevermind, prek is absolutely susceptible!). It would be really nice to have a utility to verify the hash and tag match. Since this would need the repo, I think it would make sense in `prek util`, but it could be a flag on `prek validate-config` instead. If it is in `util`, maybe it could even update the frozen tag if it's wrong.

Though I'd (also?) want a pre-commit hook tool too, also looking into that separately.

For example, `repo: https://github.com/henryiii/cibuildwheel rev: 58a0b274ea29c1e7899d45ab324b4ccdfc78d17d`, which is _not_ my commit, is happy downloaded by prek and pre-commit - it fails due to no hook file, but check the readme of that repo downloaded by either one, it's `# i like spam`!.

j178:
Detect pinned commits without lazy fetching during auto-update.

- check commit presence with `git --no-lazy-fetch cat-file -e`
- avoid treating branch-only or GitHub impostor commits as present

Closes #1864

### DIFF
--- a/crates/prek/src/cli/auto_update.rs
+++ b/crates/prek/src/cli/auto_update.rs
@@ -2,6 +2,7 @@ use std::fmt::{self, Write};
 use std::ops::Range;
 use std::path::{Path, PathBuf};
 use std::process::Stdio;
+use std::sync::OnceLock;
 use std::time::{SystemTime, UNIX_EPOCH};
 
 use annotate_snippets::{AnnotationKind, Level, Renderer, Snippet, renderer::DecorStyle};
@@ -11,6 +12,7 @@ use itertools::Itertools;
 use lazy_regex::regex;
 use owo_colors::OwoColorize;
 use prek_consts::PRE_COMMIT_HOOKS_YAML;
+use prek_consts::env_vars::EnvVars;
 use rustc_hash::FxHashMap;
 use rustc_hash::FxHashSet;
 use semver::Version;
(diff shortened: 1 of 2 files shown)


shaanmajid:
Unsure of the performance impact of this, but just an idea -- it would be ideal if this check could automatically happen during hook installation. Most folks probably won't manually execute this to check if they have a malicious config file, but if it's surfaced during normal workflows, everyone will reap the benefit.

j178:
@shaanmajid Good idea. I started with `auto-update` because the repo-fetching model there is quite different from installation.

During installation we currently use shallow clones (`--depth=1`), while `auto-update` uses a blob-less fetch. Detecting impostor commits reliably would likely require fetching a much larger commit graph, which would be a fairly invasive change to the installation path.

Adding the check in `auto-update` first is a more contained change, and it lets us validate the approach before expanding it further. If this works well, we can definitely explore bringing the same protection to installation later.

PR #1944, opened 2026-04-14
fix: use single space before frozen comment in YAML config

### LINKED ISSUE #1941: `auto-update --freeze` adds 2 spaces before `# frozen:` comment, causing `prettier` hook to fail
## Summary

`prek auto-update --freeze` adds 2 spaces before the `# frozen:` comment. `prettier` enforces a single space before inline `#` comments in `yaml`, so the subsequent `prek run -a` reformats the modified lines and reports the run as failed.

## Reproduction

```sh
prek run -a                  # passes, prettier formats file with 1 space
prek auto-update --freeze    # updates revs, adds 2 spaces before # frozen:
prek run -a                  # prettier hook rewrites the lines → hook fails
```

## Example output of `auto-update --freeze`

```yaml
- repo: https://github.com/rbubley/mirrors-prettier
  rev: b25652612a4a3c1da8caccfca9af661db56b2c07  # frozen: v3.8.2
#                                              ^^
```

## Expected

```yaml
rev: b25652612a4a3c1da8caccfca9af661db56b2c07 # frozen: v3.8.2
```

## Suggested fix

Use a single space when appending the frozen comment.

Ai-chan-0411:
## Summary
- Fix double-space before `# frozen:` comment in YAML config output
- The `render_updated_yaml_config()` function used `"  # frozen:"` (two spaces) instead of `" # frozen:"` (one space), inconsistent with the TOML path which correctly uses single space
- Updated corresponding test expected values to match

## Details
In `auto_update.rs`, the format string at the YAML frozen comment render had an extra leading space. The TOML path (`render_updated_toml_config`) already uses the correct single-space format.

Fixes #1941

## Test plan
- [x] Updated all 15 test assertions to use single space
- [x] Verified TOML path already uses single space (consistent)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/crates/prek/src/cli/auto_update.rs
+++ b/crates/prek/src/cli/auto_update.rs
@@ -1484,7 +1484,7 @@ fn render_updated_yaml_config(
         let new_rev = serialize_yaml_scalar(&revision.rev, &caps[3])?;
 
         let comment = if let Some(frozen) = &revision.frozen {
-            format!("  # frozen: {frozen}")
+            format!(" # frozen: {frozen}")
         } else if caps[5].trim_start().starts_with("# frozen:") {
             String::new()
         } else {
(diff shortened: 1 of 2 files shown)


j178:
Thanks, I think we should preserve the original spacing instead of fixed 1 space. I fixed it in #1945.

Issue #2006, opened 2026-04-26
Markdownlint hook fails to run with prek but works in pre-commit


mariusz-tang:
### Summary

Hi there!

The [markdownlint](https://github.com/markdownlint/markdownlint) hook fails to run with prek but works with pre-commit.

I get this error message:
```
  /home/mariusz/.cache/prek/hooks/ruby-bZ78BhS89Bkcqt4t5dzs/gems/bin/mdl: line 6: /home/mariusz/.cache/prek/hooks/ruby-bZ78BhS89Bkcqt4t5dzs/gems/bin/ruby: No such file or directory
```

That `mdl` script looks like this:
```
#!/bin/sh
# -*- ruby -*-
_=_\
=begin
bindir="${0%/*}"
exec "$bindir/ruby" "-x" "$0" "$@"
=end
#!/home/mariusz/.cache/prek/tools/ruby/3.3.0/bin/ruby
#
# This file was generated by RubyGems.
#
# The application 'mdl' is installed as part of a gem, and
# this file is here to facilitate running it.
#
...
```

Clearly it needs to use the ruby executable found at `~/.cache/prek/tools/ruby/3.3.0/bin/ruby`, which does exist, but to do that it tries to use another ruby executable which does not.

I'm not familiar with any of this so that's all I have for you. Let me know if there is anything else I can provide to help.

### Willing to submit a PR?

- [ ] Yes — I’m willing to open a PR to fix this.

### Platform

Linux 6.19.13-zen1-1-zen x86_64 GNU/Linux

### Version

prek 0.3.10 (1981c51f5 2026-04-21)

### .pre-commit-config.yaml

```yaml
---
repos:
  - repo: https://github.com/markdownlint/markdownlint
    rev: v0.15.0
    hooks:
      - id: markdownlint
```

### Log file

```
2026-04-26T23:28:35.428021Z DEBUG prek: 0.3.10 (1981c51f5 2026-04-21)
2026-04-26T23:28:35.428048Z DEBUG Args: ["prek", "--all-files"]
2026-04-26T23:28:35.429046Z TRACE get_root: close time.busy=973µs time.idle=2.99µs
2026-04-26T23:28:35.429067Z DEBUG Git root: /home/mariusz/projects/dstamp_1
2026-04-26T23:28:35.429078Z DEBUG Found workspace root at `/home/mariusz/projects/dstamp_1`
2026-04-26T23:28:35.429082Z TRACE Include selectors: ``
[...]

PR #2021, opened 2026-04-27
ruby: put resolved Ruby's bin dir on `$PATH` for `gem` invocations

xiaoyanli-lyft:
### Summary

`build_gemspec`, `resolve_gems`, `install_single_gem`, and `install_gems_sequential` in `crates/prek/src/languages/ruby/gem.rs` all spawn `<ruby> -S gem ...`. Ruby's `-S` flag searches `$PATH` for the named script.

The rv-ruby tarballs that prek auto-downloads ship `gem` next to `ruby` in the same `bin/` directory, but prek wasn't putting that directory on `$PATH` for these child commands. In environments without a system Ruby on `PATH` (e.g. a vanilla Docker image / CI container), the `-S` lookup fails and Ruby aborts with:

```
/<PREK_HOME>/tools/ruby/<version>/bin/ruby: No such file or directory -- gem (LoadError)
```

The first place this trips is `build_gemspec` (called from `Ruby::install` before `install_gems`), so the user-visible failure is `Failed to install hook ...: Failed to build gemspecs`. Once that's worked around, `gem install --explain` would fail identically — same root cause for all four call sites.

This is distinct from #2006 / #2017, which fixed missing `gems/bin/ruby` shim used at hook **run** time. The bug here is at hook **install** time and is not mitigated by 0.3.11 (`gem.rs` is unchanged across v0.3.10..v0.3.11).

Reproduction repo / context: filed as a separate issue with the full repro alongside this PR.

### Fix

Prepend the resolved Ruby's bin dir to `$PATH` for every `ruby -S gem` invocation:

- New `ruby_path_env(&RubyResult) -> Result<OsString>` helper that uses `prek_consts::prepend_paths` (same helper used by `Ruby::run`).
- `gem_env` now takes `&RubyResult` in addition to `gem_home` and sets `PATH` alongside the existing `GEM_HOME` / `BUNDLE_*` / `MAKEFLAGS` env. Returns `Result` so the `prepend_paths` join error can propagate (vanishingly rare in practice).
- `build_gemspec` doesn't have a `GEM_HOME` to set so it doesn't use `gem_env`, but it sets `PATH` directly via the same helper.

### Test plan

- `cargo fmt --check -p prek` — clean
- `cargo clippy -p prek --all-targets -- -D warnings` — clean
- `cargo test -p prek --bin prek` — 400 passed; 0 failed (existing unit tests, including all five `languages::ruby::gem::tests::*`)
- New integration test: `languages::ruby::auto_download_gem_install_without_gem_on_path` (CI-only, mirrors the existing `auto_download` gating). It:
  - pins `language_version: '3.2.9'` to force prek to auto-download Ruby via rv-ruby (and asserts a `tools/ruby/3.2.9*` directory exists after the run);
  - drops a `.gemspec` into the local repo so `build_gemspec` runs;
  - sets `additional_dependencies: ["webrick"]` so `resolve_gems` (`gem install --explain`) and `install_single_gem` (`gem install`) run;
  - scrubs `ruby` and `gem` from `$PATH` via the existing `remove_bin_from_path` helper, so `ruby -S gem ...` only resolves if the resolved Ruby's bin dir is prepended (i.e. the test would have failed before this fix with the `LoadError` quoted above).

The pre-existing `auto_download` test only ran `ruby --version` and never hit `gem`; `gemspec_workflow` used `language_version: system` and so picked `gem` up from the system `PATH`. The new test combines auto-download + gem operations + scrubbed `PATH` to lock the regression in.

### Manual verification of the failure mode (without this fix)

In a container with no system Ruby:

```yaml
default_language_version:
    ruby: 3.2.1
repos:
  - repo: https://github.com/jumanjihouse/pre-commit-hooks
    rev: 3.0.0
    hooks:
      - id: rubocop
```

```sh
prek install-hooks
# error: Failed to install hook `rubocop`
#   caused by: Failed to build gemspecs
#   caused by: Command `gem build` exited with an error:
#   /.../tools/ruby/3.2.1/bin/ruby: No such file or directory -- gem (LoadError)
```

With this patch, prek prepends `<ruby_prefix>/bin` to `$PATH`, `ruby -S gem` resolves, and the install proceeds.


j178:
Thanks!

PR #2077, opened 2026-05-13
Migrate `std::fs` to `fs_err`

nikosavola:
`fs_err` appears to be a dependency in the repo but it is not used everywhere. There should be more benefit of having it as a dependency if it is used everywhere in that case.


j178:
Thanks!

PR #2078, opened 2026-05-13
Refactor unsafe usage and improve documentation in prek-pty

nikosavola:
This pull request adds missing safety comments to usages of `unsafe` code throughout the codebase. Additionally, one unsafe section can be circumvented by using rustix, which is already a dependency.


j178:
Thanks for sharing this. I think this adds unnecessary complexity just to avoid `unsafe`, and `zero-unsafe` is not a goal for this project.

I don’t think this change is needed, but I appreciate the PR.

PR #2082, opened 2026-05-14
Remove unnecessary clones

FranciscoTGouveia:
Remove unnecessary clones when the receiver can be moved instead, since it is no longer used afterwards.

CC @nunoplopes


j178:
Thanks!

PR #2110, opened 2026-05-20
Run same-depth projects concurrently

### LINKED ISSUE #673: Run hooks from different project concurrently
Now that we’ve got workspace mode, if we’re sure two projects are completely separate (like ./project1 and ./project2), we can run their hooks at the same time.

j178:
Closes #673

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2070,6 +2070,7 @@ dependencies = [
  "cargo_metadata",
  "clap",
  "clap_complete",
+ "console",
  "ctrlc",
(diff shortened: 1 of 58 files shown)


j178:
Hey @potiuk, this is ready for testing now :). I did a quick test against the Airflow repo, and it looks like this should bring a significant speedup to the overall prek check time. However, I’m still not entirely sure about the correctness, so I’d really appreciate hearing how it works in your actual runs before I merge this.

PR #2111, opened 2026-05-20
Make rustup install profile configurable

Carlomus:
If rust is not already present on the system, `language: rust` toolchains are always installed with `rustup toolchain install --profile minimal`, which omits rustfmt and clippy and breaks possible `cargo fmt`/`cargo clippy` pre-commit hooks. Added PREK_RUST_PROFILE envvar so users can opt into `default` (rustfmt + clippy) or `complete`. Default is unchanged from `minimal`.


j178 (approved):
Thanks!

PR #2203, opened 2026-06-11
Add conda language support

### LINKED ISSUE #52: Support `conda` language

j178:
Closes #52, `pixi` support not included, we might add specific `language: pixi` support later.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -124,7 +124,7 @@ jobs:
         id: nextest
         shell: bash
         run: |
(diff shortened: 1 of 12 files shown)


baszalmstra:
Instead of using `conda` directly, I recommend you look at [`rattler`](https://github.com/conda/rattler) its a rust native implementation (which is what pixi is using). That has the advantage that you don't need any external tools! 

I can also guarantee that it will be much faster.

j178:
That may be next step, see #2204

PR #2224, opened 2026-06-18
fix: avoid duplicate lock file maintenance title

shaanmajid:
#2192 currently shows the repetetive title "Lock file maintenance Lock file maintenance". Notably, similar previous PRs (e.g., #2123) had the correct title "Lock file maintenance", with no repeated phrasing. This change appears to be a result of renovatebot/renovate#43629.


j178:
Thanks!

Issue #1745, opened 2026-03-04
Verify uv downloads and make mirror selection deterministic


shaanmajid:
## Problem

prek downloads uv by racing GitHub against PyPI and three third-party mirrors, using whichever responds first. No SHA256 verification on any source.

The mirrors exist for good reason -- GitHub is unreliable in some network environments. But the current design means the install source is determined by network latency, and a compromised or stale mirror can win silently.

## Proposed solution

`CUR_UV_VERSION` is already pinned in source. Embed per-platform SHA256 checksums alongside it. Astral publishes `sha256.sum` in every GitHub release, so checksums are available at version-bump time.

This makes mirror racing safe: download from whoever is fastest, verify against the embedded hash. Mirrors become untrusted CDNs rather than trusted sources.

The Renovate workflow that bumps `CUR_UV_VERSION` would need a `postUpgradeTask` to fetch and embed the new checksums -- a helper script or mise task.

## Alternatives considered

- Cross-origin runtime verification (fetch checksum from GitHub, binary from mirror): reintroduces the GitHub reachability problem the mirrors exist to solve.
- Same-origin checksums from each mirror: a compromised mirror controls both the binary and the checksum.
- Drop mirror racing, make mirrors opt-in only: degrades the experience for users with unreliable GitHub access.

shaanmajid:
Taking a look at this one again -- iirc the reason for branching logic was due to rate-limit concerns. Since initial implementation of uv support, Astral has shifted towards making their https://releases.astral.sh CDN as their default distribution source for metadata and artifact downloads.

[astral-sh/versions](https://github.com/astral-sh/versions) is Astral’s versions manifest repo; it publishes NDJSON metadata like `v1/uv.ndjson`, with uv versions, per-platform artifact URLs, archive formats, and SHA256s. That same metadata is mirrored at `https://releases.astral.sh/github/versions/main/v1/...`.

This is used by e.g., setup-uv, `uv self update`, and `uv'`s binary install paths for `ruff` and `ty`.

So I think the cleaner path here is to rewrite prek’s current uv bootstrap path to use a single Astral path (still retaining config for mirror URL overrides).

It should:
- resolve the uv artifact from the versions manifest
- download via `releases.astral.sh` by default
- support an Astral-mirror override
- (eventually) verify the archive before extraction (in a followup PR :D)

Thoughts @j178 ?

j178:
[...]

PR #1516, opened 2026-02-01
Add `language: deno` support

shaanmajid:
Closes #619

PR #2229, opened 2026-06-21
Verify managed toolchain downloads before installation

### LINKED ISSUE #1744: Verify SHA256 checksums for managed toolchain downloads
## Problem

prek downloads and executes toolchain binaries without verifying integrity. A corrupted CDN cache, truncated download, or tampered artifact installs silently.

Four of the five managed toolchain upstreams publish SHA256 checksums:

| Toolchain | Checksum source |
|-----------|----------------|
| Node | `SHASUMS256.txt` from `nodejs.org/dist/` |
| Bun | `SHASUMS256.txt` from GitHub release assets |
| Go | `sha256` field in `go.dev/dl/?mode=json` |
| Rustup | `.sha256` sidecar files alongside each binary |
| Ruby (rv-ruby) | None published |

## Proposed solution

Fetch the upstream checksum before download. Compute SHA256 while streaming to a temp file. Reject on mismatch before extraction. Fail closed: if the checksum is unavailable, malformed, or mismatched, abort the install.

This requires refactoring the download pipeline from streaming-into-extractor to download-then-verify-then-extract.

I have a draft PR covering Node, Bun, Go, and Rustup.

## Notes

- Ruby (rv-ruby) does not publish checksums. A self-hash fallback (hash on first download, detect changes on subsequent installs) could cover this gap separately.
- Node and Bun also publish GPG signatures for their checksum files. Verifying those would close the same-origin trust gap for those two toolchains, but requires embedding public keys and a PGP verification dependency. Worth considering as a follow-up.

shaanmajid:
## Summary

Adds SHA-256 verification for managed runtime downloads before extraction or installation.

Covered runtimes:

- Node
- Bun
- Go
- Rustup
- Ruby
- Deno

`uv` is intentionally not covered here, pending resolution of #1745.

Closes #1744 

## Details

This adds shared checksum-verified download helpers and wires them into the managed runtime installers. The installer now downloads the archive or installer into a temporary file, verifies it against the checksum published by the runtime’s distribution source, and only extracts or installs after the digest matches.

Checksum sources:

- Node: `SHASUMS256.txt` from `nodejs.org`
- Bun: release `SHASUMS256.txt`
- Go: `sha256` from Go release metadata
- Rustup: `.sha256` sidecars from `static.rust-lang.org`
- Ruby: `SHA256SUMS` from the same rv-ruby release download location as the archive
- Deno: `.zip.sha256sum` sidecars from Deno downloads

For SHA-256 hashing, this uses `aws-lc-rs`, which was already present in the dependency graph through the existing TLS stack. This PR adds it as an explicit workspace dependency so the checksum helper can use its digest API directly, instead of adding a separate SHA-256 crate such as `sha2`.

## Ruby mirrors

Ruby keeps the existing GitHub API metadata discovery behavior.

`PREK_RUBY_MIRROR` may point at private/vetted/air-gapped mirrors, but mirrors must provide compatible Ruby archive assets plus `SHA256SUMS` from the same release download location.

`GITHUB_TOKEN` is only sent for exact HTTPS GitHub repository sources. Non-GitHub Ruby mirrors are used without GitHub authentication.

## Compatibility notes

Managed Deno downloads now require releases using the current `.zip.sha256sum` sidecar naming. Deno `v2.1.0` and newer use that layout. Older releases such as `v2.0.0` used a different sidecar naming pattern, and older `1.x` releases may not be compatible with this verified managed-download path. Those versions can still be used via a separately installed system Deno.

Managed Bun downloads require the release to provide `SHASUMS256.txt`. That file exists for modern Bun releases, including `bun-v1.0.0`; very old pre-1.0 releases may fail checksum metadata lookup and should use a system Bun if needed.

## Security model

This catches corrupted or unexpectedly changed download bytes before extraction/install.

It does not add an independent trust root. If an upstream release, mirror, checksum file, or distribution channel is compromised together with the archive, SHA-256 verification alone cannot detect that. Signature verification, where applicable, would be worthwhile future hardening.

### DIFF
--- a/crates/prek/src/checksum.rs
+++ b/crates/prek/src/checksum.rs
@@ -1,117 +1,27 @@
 use std::fmt;
-use std::path::Path;
 use std::pin::Pin;
 use std::str::FromStr;
 use std::task::{Context as TaskContext, Poll};
 
 use anyhow::{Context, Result, bail};
 use aws_lc_rs::digest::{Context as Sha256Context, SHA256};
-use tokio::io::{AsyncRead, AsyncReadExt, ReadBuf};
+use tokio::io::{AsyncRead, ReadBuf};
 
 #[derive(Debug, Clone, Copy, Eq, PartialEq)]
 pub(crate) struct Sha256Digest([u8; 32]);
 
-pub(crate) struct Sha256Hasher(Sha256Context);
-
-impl Sha256Hasher {
-    pub(crate) fn new() -> Self {
-        Self(Sha256Context::new(&SHA256))
-    }
-
-    pub(crate) fn update(&mut self, data: &[u8]) {
-        Sha256Context::update(&mut self.0, data);
-    }
-
-    pub(crate) fn finish(self) -> Sha256Digest {
-        let digest = self.0.finish();
-        Sha256Digest::from_bytes(digest.as_ref())
-    }
-}
-
-pub(crate) struct HashReader<R> {
-    inner: R,
-    hasher: Sha256Hasher,
-}
-
-impl<R> HashReader<R> {
-    pub(crate) fn new(inner: R) -> Self {
-        Self {
-            inner,
-            hasher: Sha256Hasher::new(),
-        }
-    }
-
-    pub(crate) fn finish(self) -> Sha256Digest {
-        self.hasher.finish()
-    }
-}
-
-impl<R: AsyncRead + Unpin> AsyncRead for HashReader<R> {
-    fn poll_read(
-        self: Pin<&mut Self>,
-        cx: &mut TaskContext<'_>,
-        buf: &mut ReadBuf<'_>,
-    ) -> Poll<std::io::Result<()>> {
-        let this = self.get_mut();
-        let filled_before = buf.filled().len();
-
-        match Pin::new(&mut this.inner).poll_read(cx, buf) {
-            Poll::Ready(Ok(())) => {
-                let filled = buf.filled();
-                if filled.len() > filled_before {
-                    this.hasher.update(&filled[filled_before..]);
-                }
-                Poll::Ready(Ok(()))
-            }
-            other => other,
-        }
-    }
-}
-
 impl Sha256Digest {
+    fn from_bytes(bytes: &[u8]) -> Self {
+        let mut digest = [0_u8; 32];
+        digest.copy_from_slice(bytes);
+        Self(digest)
+    }
+
     pub(crate) fn verify(self, actual: Self, subject: impl fmt::Display) -> Result<()> {
         if actual != self {
             bail!("SHA256 checksum mismatch for `{subject}`: expected {self}, got {actual}");
         }
         Ok(())
-    }
-
-    pub(crate) async fn verify_file(self, path: &Path) -> Result<()> {
-        let actual = Self::from_file(path).await?;
-        self.verify(actual, path.display())
-    }
-
-    async fn from_file(path: &Path) -> Result<Self> {
-        let mut file = fs_err::tokio::File::open(path).await.with_context(|| {
-            format!(
-                "Failed to open `{}` for SHA256 verification",
-                path.display()
-            )
-        })?;
-        let mut hasher = Sha256Context::new(&SHA256);
-        let mut buf = [0_u8; 8192];
-
-        loop {
-            let read = file.read(&mut buf).await.with_context(|| {
-                format!(
-                    "Failed to read `{}` for SHA256 verification",
-                    path.display()
-                )
-            })?;
-            if read == 0 {
-                break;
-            }
-            Sha256Context::update(&mut hasher, &buf[..read]);
-        }
-
-        let digest = hasher.finish();
-        Ok(Self::from_bytes(digest.as_ref()))
-    }
-
-    fn from_bytes(bytes: &[u8]) -> Self {
-        let mut result = [0_u8; 32];
-        result.copy_from_slice(bytes);
-        Self(result)
     }
 }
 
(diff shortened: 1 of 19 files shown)


crates/prek/src/checksum.rs
```diff
@@ -0,0 +1,159 @@
...
+        let line = line.trim();
+        if line.is_empty() || line.starts_with('#') {
+            continue;
+        }
+
+        let Some((digest, name)) = line.split_once(char::is_whitespace) else {
+            continue;
+        };
+        let name = name.trim().trim_start_matches('*');
+        if name == filename {
```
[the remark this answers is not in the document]
shaanmajid: This seems incorrect. Bun’s current `SHASUMS256.txt` is still line-oriented; each line is `<sha256>  <filename>`. I checked the latest release (`bun-v1.3.14`) and `bun-v1.0.0`; both match the parser’s expectation. See e.g., https://github.com/oven-sh/bun/releases/download/bun-v1.3.14/SHASUMS256.txt

Also, CI exercises a managed Bun download path which forces prek to download managed Bun 1.2 and run it. If this parser failed to find `bun-*.zip`, that test should fail before the hook runs.

crates/prek/src/languages/deno/installer.rs
```diff
@@ -223,44 +224,70 @@ impl DenoInstaller {
...
-            #[cfg(unix)]
-            {
-                use std::os::unix::fs::PermissionsExt;
-                let mut perms = fs_err::tokio::metadata(&target_binary).await?.permissions();
-                perms.set_mode(0o755);
-                fs_err::tokio::set_permissions(&target_binary, perms).await?;
-            }
-
-            anyhow::Ok(())
+        let checksum = self.fetch_checksum(&url, &filename).await?;
```
j178: I’d prefer to make missing checksum metadata non-fatal for now.

If we have an expected digest and the downloaded bytes don’t match, that should absolutely be a hard error. But if the upstream release doesn’t publish a checksum file, or the checksum file simply doesn’t list this artifact, I think we should warn and continue with the existing install path.

That gives us the security win where checksum data exists, without making checksum coverage a new compatibility requirement on day one. It also avoids breaking older upstream releases or mirrors that don’t provide checksum assets yet.

crates/prek/src/checksum.rs
```diff
@@ -0,0 +1,159 @@
+use std::fmt;
+use std::path::Path;
+use std::str::FromStr;
+
+use anyhow::{Context, Result, bail};
+use aws_lc_rs::digest::{Context as Sha256Context, SHA256};
+use tokio::io::AsyncReadExt;
+
+#[derive(Debug, Clone, Copy, Eq, PartialEq)]
+pub(crate) struct Sha256Digest([u8; 32]);
```
j178, line 11: Could we add a small `HashReader<R: AsyncRead>` and wrap the reqwest response body with it?

Since the existing extraction path already works with `AsyncRead`, we could keep the download streaming and update the SHA-256 context as bytes are read. At EOF, the reader can expose the computed digest for comparison.

crates/prek/src/checksum.rs
```diff
@@ -0,0 +1,159 @@
...
+    for line in contents.lines() {
+        let line = line.trim();
+        if line.is_empty() || line.starts_with('#') {
+            continue;
+        }
+
+        let Some((digest, name)) = line.split_once(char::is_whitespace) else {
+            continue;
+        };
+        let name = name.trim().trim_start_matches('*');
```
j178: Why strips `*` here?

j178 (commented):
Thank you for working on this!

crates/prek/src/checksum.rs
shaanmajid: Good call, done!

crates/prek/src/checksum.rs
shaanmajid: Updated to tighten the regex a bit and added an explanatory comment, but standard checksum formatting may include a single `*` character for binary-mode files.

crates/prek/src/languages/deno/installer.rs
shaanmajid: I’m slightly hesitant to make missing checksum metadata non-fatal by default, since I think that materially undermines the security posture here.

Regarding releases, I compiled checksum availability for all supported toolchains. The concrete gaps I found are mostly old releases:

- Bun: `v0.1.1` through `v0.5.0` have downloadable `bun-linux-x64.zip` assets but no `SHASUMS256.txt`; `v0.6.0+` (released May 16, 2023) sampled has checksum metadata.
- Deno: `v1.x` and `v2.0.0` have downloadable archives but no `.sha256sum`; `v2.0.1+` (released October 17, 2024) sampled has checksum metadata. (For reference, Deno v1 was released May 13, 2020).
- Node, Go, and rustup look covered for the downloadable artifacts I sampled.
- rv-ruby’s default release has `SHA256SUMS`; custom `PREK_RUBY_MIRROR` is the main place where this could become a mirror compatibility issue.

Although some old Bun and Deno versions are missing checksums, I think dropping support for both are justified here, since (1) those are relatively new toolchains in general, and (2) support for both is prek-only (not pre-commit) and only added within the last few months in #1411 and #1516. Because of both, I doubt people are installing old versions.

Given that, I think fail-closed is defensible for official managed downloads -- if an official source does not publish checksum metadata for an artifact at install-time (despite currently publishing this metadata today), then installing it silently weakens the security guarantee from this change. I'm concerned that if the only flag is e.g., a warning, users will likely ignore it, especially in CI.

My only legitimate worry of checksum availability is for consumers of Ruby mirrors via the `PREK_RUBY_MIRROR` environment variable; I'm not sure how often this feature is used (if at all).

If necessary, would you be open to a narrower compatibility exception instead? E.g., switch to warn-and-continue only for `PREK_RUBY_MIRROR` / custom mirrors, or only behind an explicit environment variable?

If you still prefer warn-and-continue everywhere where the checksum is missing for the first release, I can make that change, but just wanted to call out that I don't think this case should ever be hit with current usage (only exception being the Ruby mirror point I noted above).

crates/prek/src/languages/deno/installer.rs
```diff
@@ -223,44 +224,70 @@ impl DenoInstaller {
...
+            Self::install_extracted(&target, extracted).await
         })
         .await
         .context("Failed to download and extract deno")?;
 
         Ok(DenoResult::from_dir(&target).with_version(version.clone()))
     }
 
+    async fn fetch_checksum(&self, url: &str, filename: &str) -> Result<Sha256Digest> {
+        let checksum_url = format!("{url}.sha256sum");
```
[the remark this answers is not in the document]
shaanmajid: This seems wrong; yes, the GitHub release assets publish checksums, but the `dl.deno.land` endpoint does do, e.g., https://dl.deno.land/release/v2.8.3/deno-x86_64-unknown-linux-gnu.zip.sha256sum

I chose to fetch from `dl.deno.land` instead of GitHub release assets since that is the existing artifact download source.

crates/prek/src/languages/deno/installer.rs
j178: Thanks for digging into the availability data.

I’m still uncomfortable making missing checksum metadata fatal by default in this version. Deno is the main example: Deno releases before 2.0 don’t have checksum sidecars, so fail-closed would make those versions uninstallable in prek. That is a breaking compatibility change, and it also suggests checksum sidecars are not yet a universal enough convention to make them a hard requirement for every managed toolchain we add later.

I pushed 309d343ce7badb411f624a754ebbb629714c590b, which makes this policy explicit via `PREK_DOWNLOAD_CHECKSUM_POLICY`.

The default is `warn-missing`: checksum mismatch remains a hard error, but missing checksum metadata warns and continues. Users who want fail-closed behavior can set `PREK_DOWNLOAD_CHECKSUM_POLICY=required`, and users who need to bypass checksum lookup entirely can set `disabled`.

That keeps the security win when metadata exists, while avoiding a compatibility break for older releases or mirrors that don’t ship sidecars yet.

PR #2255, opened 2026-06-26
Skip unnecessary submodule updates during hook repo clone

shaanmajid:
## Summary

Skip `git submodule update --init --recursive` after cloning a hook repo when the checked-out repo has neither `.gitmodules` nor gitlink entries.

Most hook repositories do not use submodules. Before this change, `prek` still ran `git submodule update` for every cloned remote repo, which adds an avoidable Git subprocess on cold cache paths.

The new behavior keeps the existing submodule path when either:

- `.gitmodules` exists
- the checked-out index contains gitlinks (`160000` entries)

That preserves behavior for normal submodule repos, broken `.gitmodules` cases, and gitlink-without-`.gitmodules` edge cases.

## Benchmarks

Benchmarked cold `prek list` with fresh isolated `PREK_HOME` against local mirrors of real hook repositories to avoid network variance. Baseline was `upstream/master`; changed was this branch. Each scenario used 40 runs and 6 warmups.

| Repo | Rev | Files | Baseline | Changed | Result |
| --- | --- | ---: | ---: | ---: | ---: |
| `pre-commit/pre-commit-hooks` | `v6.0.0` | 106 | `132.3 ± 12.6 ms` | `84.5 ± 11.0 ms` | `1.57x faster` |
| `astral-sh/ruff-pre-commit` | `v0.9.9` | 10 | `113.4 ± 14.2 ms` | `76.8 ± 7.1 ms` | `1.48x faster` |
| `psf/black` | `25.1.0` | 424 | `274.7 ± 24.9 ms` | `231.6 ± 18.9 ms` | `1.19x faster` |

All benchmarked repos had no `.gitmodules` and no gitlinks. Trace logs confirmed the old path ran:

```bash
git submodule update --init --recursive --depth=1
```

The new path checked gitlinks with:

```bash
git ls-files -z -s
```

and skipped submodule update.


j178:
Thank you, a very good improvement!

PR #2260, opened 2026-06-28
Add Nix flake for installation via nix run/profile install

levonk:
## What

Adds a `flake.nix` so prek can be installed and run directly from GitHub at the latest release:

```bash
nix run github:j178/prek
nix profile install github:j178/prek
nix run github:j178/prek/v0.4.5   # pin a specific release
```

## Why

prek is already packaged in nixpkgs, but the nixpkgs unstable channel lags behind upstream releases (currently shipping 0.3.11 vs the latest 0.4.5). A flake in this repo gives users the current release directly from GitHub without waiting for nixpkgs to catch up, and provides:

- **One-command install / run** — `nix run github:j178/prek` with no clone or manual build steps.
- **Pure / Hermetic builds** — every input (the prebuilt binary, glibc/libiconv) is pinned in `flake.lock`. If it builds today, it builds in ten years.
- **Reproducible** — the exact same derivation always produces the exact same output bit-for-bit. No "works on my machine."
- **Idempotent installs** — running `nix profile install` twice is a no-op.
- **Rollback-able** — `nix profile rollback` restores the previous profile generation instantly.
- **Cross-platform** — same invocation on macOS (Apple Silicon & Intel) and Linux. The flake handles platform-specific linking.
- **Atomic upgrades / downgrades** — profiles are switched atomically. No half-upgraded state.
- **Clean uninstall** — `nix profile remove` leaves no residue.

## Changes

- `flake.nix`: Nix flake wrapping the v0.4.5 prebuilt release tarball as `packages.<system>.default` and `apps.<system>.default` for x86_64/aarch64 linux+darwin. Uses `autoPatchelfHook` on Linux for glibc linking.
- `flake.lock`: pinned `nixpkgs-unstable` input.
- `.gitignore`: added `/result` and `/result-*` Nix build symlinks.
- `README.md`: extended the existing Nix install section (`<!-- --8<-- [start: nix-install] -->`) with flake-based install/run commands alongside the existing nixpkgs path. `docs/installation.md` picks this up automatically via its `--8<-- "README.md:nix-install"` include.

## Testing

Verified locally on `x86_64-darwin`:

```bash
nix flake check --no-build   # passed
nix build .                  # passed
result/bin/prek --version    # prek 0.4.5 (bb3108283 2026-06-15)
nix run . -- --help          # passed
```

No source files in `crates/` are touched — the diff is `flake.nix`, `flake.lock`, `.gitignore`, and a README snippet, so the existing test suite is unaffected.

[...]

PR #2279, opened 2026-06-30
feat: add Nix flake for installation via nix run/profile install

levonk:
## What

Adds a `flake.nix` so prek can be installed and run directly from GitHub at the latest release:

```bash
nix run github:j178/prek
nix profile install github:j178/prek
```

The flake tracks `master` and is auto-bumped to the latest release by a daily workflow, so `github:j178/prek` always serves the current release.

## Addressing the maintenance concern (re: #2260)

@j178's objection to #2260 was that it added per-release maintenance — every release would need `version` and four `sha256` hashes updated by hand, with no automation. This PR adds that automation.

**New in this PR: `.github/workflows/nix-release.yml`** — a GitHub Action that:

1. Runs daily (and on manual `workflow_dispatch`) and compares `flake.nix`'s pinned `version` to the latest GitHub release.
2. When they differ, prefetches each platform asset's SRI hash via `nix store prefetch-file --json --hash-type sha256`.
3. Rewrites `version` and the per-platform `file`/`sha256` in `flake.nix` with an inline Python script.
4. Opens a PR (`chore(nix): bump flake to vX.Y.Z`) with the 5-line diff.

The maintainer cuts a release exactly as today; within 24h the workflow opens a PR with the bumped `flake.nix`. Reviewing a 5-line diff (version + 4 hashes) needs no Nix knowledge. Merge → `nix run github:j178/prek` serves the new release. **Zero manual flake editing, zero Nix knowledge required.**

### Why scheduled, not `release: published`

prek's releases are created with `secrets.GITHUB_TOKEN` (`.github/workflows/release.yml`, line 267/299 — `gh release create` with `GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}`). GitHub deliberately does not start new workflow runs from events authored by `GITHUB_TOKEN` (to prevent recursive loops), so a `release: published` trigger would **silently never fire**. A daily lag-check is fully decoupled from how releases are created, needs no PAT, and requires no edits to the cargo-dist release pipeline. It also self-heals: if a run is missed, the next day's run catches up.

The `ASSET_MAP` is the only project-specific input (four `system|asset-substring` lines). Sibling `.sha256` checksum files are filtered out automatically so substrings match the tarball, not its checksum companion.

### Why no tag-pinning (`github:j178/prek/vX.Y.Z`)

Release tags are cut by cargo-dist *before* the lag-check workflow bumps `flake.nix` on `master`. So a tag ref either lacks the flake entirely (tags before this PR merges) or contains the *previous* version's hashes (tags after merge, since the tag captures `master` before the bump PR lands). Tag-pinning is fundamentally incompatible with a post-release bump workflow. The flake serves the current release via `github:j178/prek` (tracks `master`, always has the latest bumped version). For reproducibility, pin to a specific commit SHA after a bump PR merges, or use the nixpkgs package.

## Why

prek is already packaged in nixpkgs, but the nixpkgs unstable channel lags behind upstream releases (currently shipping 0.3.11 vs the latest 0.4.5). A flake in this repo gives users the current release directly from GitHub without waiting for nixpkgs to catch up, and provides:

- **One-command install / run** — `nix run github:j178/prek` with no clone or manual build steps.
- **Pure / Hermetic builds** — every input (the prebuilt binary, glibc/libiconv) is pinned in `flake.lock`. If it builds today, it builds in ten years.
- **Reproducible** — the exact same derivation always produces the exact same output bit-for-bit. No "works on my machine."
- **Idempotent installs** — running `nix profile install` twice is a no-op.
- **Rollback-able** — `nix profile rollback` restores the previous profile generation instantly.
- **Cross-platform** — same invocation on macOS (Apple Silicon & Intel) and Linux. The flake handles platform-specific linking.
- **Atomic upgrades / downgrades** — profiles are switched atomically. No half-upgraded state.
- **Clean uninstall** — `nix profile remove` leaves no residue.

## Changes

- `flake.nix`: Nix flake wrapping the v0.4.5 prebuilt release tarball as `packages.<system>.default` and `apps.<system>.default` for x86_64/aarch64 linux+darwin. Uses `autoPatchelfHook` on Linux for glibc linking.
- `flake.lock`: pinned `nixpkgs-unstable` input.
- `.github/workflows/nix-release.yml`: **scheduled lag-check automation** that auto-bumps `version` + per-platform `sha256` hashes and opens a PR when `flake.nix` falls behind the latest release — no manual flake maintenance, no PAT, no release-pipeline edits.
- `.gitignore`: added `/result` and `/result-*` Nix build symlinks.
- `README.md`: extended the existing Nix install section (`<!-- --8<-- [start: nix-install] -->`) with flake-based install/run commands alongside the existing nixpkgs path. `docs/installation.md` picks this up automatically via its `--8<-- "README.md:nix-install"` include.

## Testing

Verified locally on `x86_64-darwin`:

```bash
nix flake check --no-build   # passed
nix build .                  # passed
result/bin/prek --version    # prek 0.4.5 (bb3108283 2026-06-15)
nix run . -- --help          # passed
```

The workflow YAML validates and its lag-check logic was dry-run against the actual v0.4.5 release: `version` comparison and asset-substring matching (with `.sha256` filtering) confirmed for all four systems. The workflow correctly reports "up to date" when `flake.nix` matches the latest release, and would set `LAGGING=yes` + open a PR when they diverge.

No source files in `crates/` are touched — the diff is `flake.nix`, `flake.lock`, `.github/workflows/nix-release.yml`, `.gitignore`, and a README snippet, so the existing test suite is unaffected.

## Notes

- The flake wraps the **prebuilt release tarball** (preferred path per the upstream release artifacts) rather than building from source, so it stays in sync with `cargo-dist` releases and avoids a full Rust toolchain in the closure.
- Only the 4 standard Nix systems are exposed; upstream also ships musl/arm/armv7/riscv64/s390x tarballs but those map to no stdenv system. To expose more, add entries to `assets` in `flake.nix` and a matching line to the workflow's `ASSET_MAP`.
- Actions are pinned to commit SHAs to match the repo's existing workflow conventions.
- PRs opened by `GITHUB_TOKEN` do not trigger downstream CI workflows; the bump PR is a 5-line hash change with no source impact, safe to merge without checks.
- No breaking changes to existing build paths.

## Scope

The PR scope is well-contained — additive only, no existing functionality affected.

## Related

Supersedes #2260 (closed) — adds the release-triggered hash automation that was missing there.

### DIFF
--- a/.github/workflows/nix-release.yml
+++ b/.github/workflows/nix-release.yml
@@ -1,0 +1,129 @@
+name: "Update Nix flake"
+
+# Checks whether `flake.nix` lags behind the latest GitHub release. If it does,
+# prefetches the new release's per-platform SRI hashes, rewrites `flake.nix`,
+# and opens a PR.
+#
+# This runs on a schedule instead of `release: published` because prek's releases
+# are created with `secrets.GITHUB_TOKEN` (see .github/workflows/release.yml),
+# and GitHub does not start new workflow runs from events created by
+# GITHUB_TOKEN — so a `release: published` trigger would never fire. A daily
+# lag-check is fully decoupled from how releases are created and needs no PAT.
+# Run manually any time via workflow_dispatch.
+
+on:
+  schedule:
+    - cron: "17 6 * * *"
+  workflow_dispatch:
+
+permissions:
+  contents: write
+  pull-requests: write
+
+concurrency:
+  group: nix-flake-release
+  cancel-in-progress: true
+
+jobs:
+  update-flake:
+    name: Bump flake version + hashes if lagging
+    runs-on: ubuntu-latest
+    if: github.repository == 'j178/prek'
+    steps:
+      - name: Checkout
+        uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+        with:
+          persist-credentials: false
+
+      - name: Install Nix
+        uses: cachix/install-nix-action@8aa03977d8d733052d78f4e008a241fd1dbf36b3 # v31.10.6
+
+      - name: Check for lag and rewrite flake.nix
+        env:
+          # system|asset-substring — one per line. The substring must uniquely
+          # match the release asset filename for that system (including the
+          # .tar.gz suffix so it does not match the sibling .sha256 files).
+          ASSET_MAP: |
+            x86_64-linux|x86_64-unknown-linux-gnu.tar.gz
+            aarch64-linux|aarch64-unknown-linux-gnu.tar.gz
+            x86_64-darwin|x86_64-apple-darwin.tar.gz
+            aarch64-darwin|aarch64-apple-darwin.tar.gz
+        run: |
+          set -euo pipefail
+          # Latest release tag from the GitHub API (strip leading 'v').
+          tag=$(curl -fsSL -H "Accept: application/vnd.github+json" \
+            "https://api.github.com/repos/${GITHUB_REPOSITORY}/releases/latest" \
+            | python3 -c 'import json,sys; print(json.load(sys.stdin)["tag_name"])')
+          latest="${tag#v}"
+          # Version currently pinned in flake.nix.
+          current=$(python3 -c 'import re; s=open("flake.nix").read(); m=re.search(r"version = \"([^\"]*)\";", s); print(m.group(1))')
+          echo "flake.nix version: $current  |  latest release: $latest (tag $tag)"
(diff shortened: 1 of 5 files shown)


j178:
Thanks for putting this together. I'm open to adding a `flake.nix` to this repo. But I still need more time to get familiar enough with Nix and this workflow before I can responsibly merge and maintain it, so I won't merge it immediately. I'll keep this open while I review it.

levonk:
Thanks for your contribution to making this open source and consideration for the patch.

PR #2281, opened 2026-07-01
chore: lock macports portfile updater

shaanmajid:
Seems worthwhile to lock a release-related script.

(This appeared to be the only PEP 723 script with actual dependencies; most others had a Python version specifier.)

Note renovate will not track this; see renovatebot/renovate#33591


j178:
Thanks!

PR #2282, opened 2026-07-01
chore: update `pyproject.toml` metadata

shaanmajid:
Updates `Development Status` classifier -- I think prek has been considered stable for some time :)

Also small cleanup to update rooster rev to the full commit SHA instead of SHA-1; I don't think this was a real concern since the `uv.lock` file had the full SHA, but seems worthwhile regardless. Using SHA-1 for anything has been generally advised against for some time now.


j178:
Thanks!

PR #2283, opened 2026-07-01
chore: add uv pre-commit hook

shaanmajid:
Seems worthwhile since we have a `uv.lock` file :)

Chose 0.11.24, since it's most recent release in compliance with 7-day cooldown.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -14,6 +14,11 @@ repos:
       - id: check-toml
       - id: end-of-file-fixer
 
+  - repo: https://github.com/astral-sh/uv-pre-commit
+    rev: 73c2d77a42a113aee9e4b748c24937f09557b82d  # frozen: 0.11.24
+    hooks:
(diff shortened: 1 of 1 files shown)


j178:
Thanks, but I may pass on this one for now. We don’t have many Python dependencies to maintain, and uv is a pretty large dependency to pull in as a pre-commit hook.

shaanmajid:
Fair enough :p

If you're open to moving this to CI, uv has a [`check-lock.yml`](https://github.com/astral-sh/uv/actions/workflows/check-lock.yml) workflow, which checks both project and PEP 723 script lockfiles; I could adapt this PR to do something similar. (Last run on main appeared to take 17s on a warm cache). 

Otherwise I'll just go ahead and close!

j178:
I think we can close this for now. A dedicated `check-lock.yml` feels like more process than we need at the moment, given how small and infrequently changing the Python dependency surface is here.

PR #2331, opened 2026-07-12
Add aliases for hook priorities

chrisoro:
In projects with lots of hooks (and therefore repos) I found it very hard to follow which hooks run in which priority. I want to run as many as possible in parallel and therefore assign them to groups. Until now, I had to assign the same integer to hooks that can run at the same time (=belong to the same group), so I always wished for this feature.

I tried to do it in such a way that needs minimal changes and decided to basically only change the config parsing. In the end, it will still result in an u32 and all further logic still applies.

I also changed validate_group_name to validate_config_name to apply the same restrictions to the priorities names as for groups.

Used AI to find all docs that needs to be updated and updated them.

My first contribution here, so tell me if I missed something or if you want some changes.

### DIFF
--- a/crates/prek/src/cli/run/selector.rs
+++ b/crates/prek/src/cli/run/selector.rs
@@ -3,7 +3,7 @@ use std::fmt::Display;
 use std::path::{Path, PathBuf};
 use std::sync::{Arc, Mutex};
 
-use crate::config::validate_group_name;
+use crate::config::validate_name;
 use crate::hook::Hook;
 use crate::warn_user;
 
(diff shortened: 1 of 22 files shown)


j178:
Thanks! I’m concerned that named priorities would introduce a second grouping concept alongside groups, but with unrelated semantics: groups select hooks, while priorities schedule them. The same natural names, such as `format`, `checks`, or `tests`, are likely to be used for both, which could be confusing.  

YAML anchors can be a workaround before we add named priorities:

```yaml
- id: ruff
  priority: &checks 10

- id: mypy
  priority: *checks
```

chrisoro:
I agree but we fully embraced prek dropping pre-commit compatibility and switched to prek.toml. There are no anchors in toml.

I see your point but selection and order are ultimately two different things. Do you think this is just a naming problem? We could also enforce the priorities to all start/end with a specific prefix/suffix to make it clearer.

Imo this would be a great feature since assigning and maintaining int based priorities in a toml file with 50 hooks is a real pain

j178:
Thanks!

PR #2353, opened 2026-07-15
Print hook description on failure with --verbose

GhostofGoes:
## Summary

When `--verbose` (or the new `PREK_VERBOSE` environment variable) is enabled and a hook fails, prek now prints the hook's `description` field alongside the existing hook id/duration/exit code details. The goal is to help users understand why a hook failed without needing to dig through `prek.toml`/`.pre-commit-config.yaml`. This is opt-in, but could become default behavior in the future.

The `PREK_VERBOSE` env var lets this work for hooks triggered via Git (e.g. `git commit`) too, since Git has no way to forward CLI flags like `--verbose` to the installed hook shim.

### Changes

- `crates/prek/src/cli/run/run.rs`: print `- description: ...` after `- hook id: ...` when a hook fails and verbose mode (global `--verbose`/`-v` or per-hook `verbose: true`) is active.
- `crates/prek-consts/src/env_vars.rs`: added `PREK_VERBOSE` env var, mirroring the existing `PREK_QUIET`.
- `crates/prek/src/cli/mod.rs`: wired `env = EnvVars::PREK_VERBOSE` onto the `--verbose` arg and updated its help text.
- `crates/prek/src/cli/hook_impl.rs` / `crates/prek/src/main.rs`: threaded `verbose` through `hook-impl` so hooks invoked by Git shims respect it.
- `crates/prek/src/config.rs`: updated the `description` field doc comment to mention it's now surfaced on failure.
- Docs: `docs/usage.md`, `docs/authoring-hooks.md`, `docs/reference/configuration.md`, `docs/reference/cli.md` (regenerated), and `prek.schema.json` (regenerated).

### Tests

- `crates/prek/tests/run.rs`: `verbose_prints_hook_description_on_failure` and `prek_verbose_env_var_prints_hook_description_on_failure`.
- `crates/prek/tests/hook_impl.rs`: `hook_impl_verbose_env_var_prints_hook_description`, exercising the Git-hook (`git commit`) path end-to-end.

`cargo fmt --check` and `cargo clippy --all-targets` are clean, and the full `run`/`hook_impl` integration test suites pass locally.

### DIFF
--- a/crates/prek-consts/src/env_vars.rs
+++ b/crates/prek-consts/src/env_vars.rs
@@ -99,6 +99,7 @@ impl EnvVars {
     pub const PREK_CONTAINER_RUNTIME: &'static str = "PREK_CONTAINER_RUNTIME";
     pub const PREK_DOCKER_NO_INIT: &'static str = "PREK_DOCKER_NO_INIT";
     pub const PREK_QUIET: &'static str = "PREK_QUIET";
+    pub const PREK_VERBOSE: &'static str = "PREK_VERBOSE";
 
     // PREK internal environment variables
     pub const PREK_INTERNAL__TEST_DIR: &'static str = "PREK_INTERNAL__TEST_DIR";
--- a/crates/prek/src/cli/hook_impl.rs
+++ b/crates/prek/src/cli/hook_impl.rs
@@ -33,6 +33,7 @@ pub(crate) async fn hook_impl(
     skip_on_missing_config: bool,
     script_version: Option<usize>,
     args: Vec<OsString>,
+    verbose: bool,
     printer: Printer,
 ) -> Result<ExitStatus> {
     let stdin = read_hook_stdin(hook_type).await?;
@@ -138,7 +139,7 @@ pub(crate) async fn hook_impl(
         false,
         false,
         run_args.extra,
-        false,
+        verbose,
         printer,
     )
     .await?;
--- a/crates/prek/src/cli/mod.rs
+++ b/crates/prek/src/cli/mod.rs
@@ -208,7 +208,10 @@ pub(crate) struct GlobalArgs {
     pub quiet: u8,
 
     /// Use verbose output.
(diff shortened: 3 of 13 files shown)


GhostofGoes:
@j178 Howdy, any feedback on this PR? Does a feature like this align with your vision for prek?

j178:
Thanks! I can see some value for hooks that fail without useful output, but `description` generally explains what a hook does rather than why a particular run failed. I’m also not convinced we should add a new global `PREK_VERBOSE` interface and thread it through `hook-impl` solely for this. I’d prefer to keep this smaller or wait for a concrete user case.

j178:
Since maintainer edits aren’t enabled on this PR, I opened #2490 with you credited as a co-author. Thanks for the contribution!

GhostofGoes:
> Since maintainer edits aren’t enabled on this PR, I opened #2490 with you credited as a co-author. Thanks for the contribution!

Awesome, thank you! 

Regarding the concrete use case, we are rolling out better linting (prek is awesome), and some junior developers are confused about why certain hooks failed. For example, invalid branch names. The current message says you're not allowed to commit to the branch, without a reason:

```shell
branch name validation...................................................Failed
- hook id: no-commit-to-branch
- exit code: 1

  You are not allowed to commit to branch 'invalid/branch-name'
```

The description, however, explains the reason:

```toml
[[repos.hooks]]
id = "no-commit-to-branch"
name = "branch name validation"
description = "Prevent committing to 'main', 'master', or any branch that aren't kebab-case (only allows letters, numbers, and hyphens)"
...
```

Sure, they could grep `prek.toml`, but a lot of users don't know how all the tooling works and where to look (not knowing what they don't know), beyond "oh I committed and the linter failed ugh". With the description, they can obviously see they should not have `/` characters.

Ideal world there would be a way for this particular hook to point out WHERE the user messed up, but that's much more complex, and this was a simple workaround 😄 .

(Ironically, the branch name for this PR violates said rules 😅 )

PR #2431, opened 2026-07-28
Do not shuffle file list for verbose output

### LINKED ISSUE #2430: File list printed with `--dry-run --verbose` is unsorted
### Before submitting

- [x] I have searched the issue list for similar bug reports.

### Summary

When running a hook with `--dry-run --verbose`, prek helpfully prints the list of files that would be checked by the hook(s). However, this list is unsorted. The output would be more readable if it was sorted.

### Platform

Darwin 25.5.0 arm64 Darwin

### Version (current: 0.4.11)

prek 0.4.11 (Homebrew 2026-07-24)

### .pre-commit-config.yaml

n/a

### Log file

n/a

allanlewis:
Fixes #2430.

The file list usually comes from `git ls-files`, which is already ordered. Preserve that order for `--dry-run --verbose` output instead of shuffling it.

### DIFF
--- a/crates/prek/src/cli/run/run.rs
+++ b/crates/prek/src/cli/run/run.rs
@@ -1389,7 +1389,9 @@ fn dry_run_hook(hook: &InstalledHook, input: &HookRunInput<'_>) -> Result<Vec<u8
 
     match input {
         HookRunInput::Filenames(filenames) => {
-            for filename in filenames {
+            let mut filenames = filenames.clone();
+            filenames.sort();
+            for filename in &filenames {
                 writeln!(output, "- {}", filename.display())?;
             }
         }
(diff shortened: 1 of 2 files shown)


crates/prek/src/cli/run/run.rs
```diff
@@ -1335,11 +1335,11 @@ async fn run_hook(
         return Ok(RunResult::from_status(hook, RunStatus::NoFiles));
     }
     let start = std::time::Instant::now();
-    input.shuffle();
 
     let hook_output = if dry_run {
         hooks::HookOutput::unchanged(0, dry_run_hook(&hook, &input)?)
     } else {
+        input.shuffle();
```
allanlewis, line 1342: This is a great solution - thanks!

Issue #1889, opened 2026-04-03
Stash conflict rollback corrupts git index, causing newly added files to be committed empty


Alex-Dufour:
## Description

When `git commit` is run via prek with unstaged changes present (in files outside the staged set), prek stashes the unstaged changes before running hooks. If a hook like `end-of-file-fixer` or `trailing-whitespace` modifies staged files, prek detects a conflict between the stash and the hook modifications, aborts the commit, and rolls back.

**The problem:** after the rollback, the git index is left in a corrupted state. Specifically, newly added files can have their content emptied in the index. If the user then re-stages and commits (especially after resolving the unstaged changes), the corrupted index content is committed — resulting in empty files being committed silently.

## Steps to reproduce

1. Have a repo with a `.pre-commit-config.yaml` that includes auto-fixing hooks (`end-of-file-fixer`, `trailing-whitespace`, etc.)
2. Create a new file (e.g., `new-file.yaml`) with content that has trailing whitespace or a missing final newline
3. Modify an existing file (e.g., `existing.py`) with similar minor formatting issues
4. Stage only a subset of the repo: `git add some-directory/` — leaving at least one modified file elsewhere unstaged
5. Run `git commit -m "feat: add new files"`

**Expected:** the commit either succeeds (with hooks auto-fixing the formatting) or fails cleanly with the index intact.

**Actual:**
- prek stashes the unstaged changes
- Hooks fix the staged files on disk
- prek detects a conflict between the stash and hook modifications
- prek rolls back and prints: `Stashed changes conflicted with changes made by hook, rolling back the hook changes`
- The commit is aborted
- **The git index is left corrupted** — newly added files may now be empty in the index (hash `e69de29b`, the empty blob)

If the user then re-stages and eventually commits, the file goes in empty.

## Terminal output from real occurrence

```
$ git add api/
$ git commit -m "feat: add email security bl for nl and be markets"
Unstaged changes detected, stashing unstaged changes to `/Users/user/.cache/prek/patches/1775206900588-36748.patch`
Files were modified by following hooks.................................................Failed
  ┌ check yaml.........................................................................Passed
  │ check json.....................................................(no files to check)Skipped
  │ check for merge conflicts..........................................................Passed
  │ fix end of files...................................................................Failed
  │ - hook id: end-of-file-fixer
  │ - exit code: 1
  │
  │ Fixing api/riskmgt/models/business_line/questionnaire_version.py
[...]

PR #2445, opened 2026-07-30
fix(run): roll back to the pre-hook tree, not the live index

CosmicTrader:
## Summary

When the stash restore conflicts, the rollback restores from the **current index** rather than the pre-hook state. If a hook has re-staged its own output, the index has already moved, so the saved patch gets applied on top of the hook's content and the user's unstaged changes are silently lost.

This is the general case behind #1889. That issue was closed because the reported scenario couldn't be reproduced from the description — this PR includes a minimal deterministic reproduction as a test.

## The mechanism

`UnstagedChangesRestorer::restore` falls back to `checkout_working_tree` when `git apply` fails, intending to discard the hook's changes and retry:

```rust
// Discard any changes made by hooks, and try applying the patch again.
Self::checkout_working_tree(&self.root)?;
Self::git_apply(patch)?;
```

`checkout_working_tree` runs `git checkout -- <root>` with no tree-ish, so it restores from the index. That's correct only while the index still holds the pre-hook state.

Auto-fixing hooks commonly break that assumption: they run a formatter and then `git add` the result, so the commit completes in one step instead of aborting for a manual re-stage. Once such a hook runs, the index holds the hook's content. The rollback then reinstates *that* as the baseline, and the patch — computed against the original content — is applied on top of it. It either fails outright or lands on the wrong lines. Either way the unstaged work is gone, and `git status` is clean afterwards, so nothing signals it.

## The fix

`clean` already computes the pre-hook tree with `git write-tree` and then discards it. This keeps it and passes it to the rollback checkout, so `git checkout <tree> -- <root>` restores both the index entries and the working tree to their true pre-hook state before the patch is re-applied.

The happy path is unchanged — the initial `clean` checkout still passes `None`, since the tree was just written from the index and the two are identical at that point. Only the conflict path differs, and it gains one argv element on a command that runs only after a conflict has already occurred.

## Reproduction

`restaging_hook_does_not_discard_unstaged_changes` in `crates/prek/tests/run.rs`. A hook rewrites a file and `git add`s it; the same file has both a staged edit and a further unstaged edit.

On master:

```
  left: "NORMALISED\nbody\nstaged edit\n"
 right: "ORIGINAL\nbody\nstaged edit\nUNSTAGED TAIL\n"
```

The unstaged tail is gone and the hook's content has replaced the user's. With the fix, the working tree comes back exactly as the user left it and the file is still reported dirty.

## Checks

- New test fails on master, passes with the change.
- `staged_files_only`, `intent_to_add_file_survives_conflicted_stash_restore`, `restore_on_interrupt`, and `all_files_with_existing_unstaged_changes_uses_snapshot_baseline` all still pass.
- Remaining failures in a full local `cargo test` run are language-toolchain tests (conda, dart, haskell, julia, lua, perl, php, r, ruby) that need toolchains not installed on this machine; they fail identically on an unmodified checkout.

Happy to adjust the approach if you'd prefer the index restored a different way.

### DIFF
--- a/crates/prek/src/cli/run/keeper.rs
+++ b/crates/prek/src/cli/run/keeper.rs
@@ -16,6 +16,15 @@ use crate::store::Store;
 struct IntentToAddRestorer(Vec<PathBuf>);
 struct UnstagedChangesRestorer {
     root: PathBuf,
+    /// Tree object written from the index BEFORE any hook ran.
+    ///
+    /// The rollback path must reset to this snapshot rather than to the live index:
+    /// hooks routinely mutate the index (any auto-fixer that re-stages what it
+    /// rewrote, e.g. a `prettier --write` + `git add` wrapper), so by the time we
+    /// roll back, `git checkout -- <root>` would restore the hook's own content as
+    /// the baseline and then apply a patch computed against the ORIGINAL content on
+    /// top of it — silently losing or duplicating unstaged work.
+    tree: String,
     patch: Option<PathBuf>,
 }
 
@@ -95,7 +104,7 @@ impl UnstagedChangesRestorer {
             .arg("--binary")
             .arg("--exit-code")
             .hidden_args(["--ignore-submodules", "--no-color", "--no-ext-diff"])
-            .arg(tree)
+            .arg(&tree)
             .arg("--")
             .arg(root)
             .check(false)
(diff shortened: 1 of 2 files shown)


CosmicTrader:
Closing — reopening this from the account that will maintain the fix. Same commit, no changes to the patch or the test. Apologies for the noise.

PR #2482, opened 2026-08-04
Silently ignore `x-` extension keys in configuration

### LINKED ISSUE #2481: Silently ignore top-level `x-` extension keys to support YAML anchors
First of all, Thank you so much for implementing `prek exec` #2475. This is extremely helpful for us. 

**Description:**
Currently, when using `x-` prefix keys to define custom fields, prek prints a "warning: Ignored unexpected keys in test_cfg.yaml: repos[0].hooks[0].x-globs, repos[0].hooks[0].x-stdin-args" during parsing. 

**Expected Behavior:**
Following the convention used by tools like [Docker Compose](https://docs.docker.com/reference/compose-file/extension/), [OpenAPI](https://swagger.io/docs/specification/v3_0/openapi-extensions/), and GitLab CI, prek should silently ignore any keys prefixed with `x-` (e.g., `x-jj-stdin-command: "<command> --stdin-filepath $path`). 

**Why?**
While I understand there might be some strong opinions in extending prek, it'd be much easier to adopt this in a large monorepo with existing users and conventions if we can add additional fields that prek can simply ignore. While it doesn't error out now, every dev seeing this warning makes it look like we configured something wrong.

**How?**
As I understand https://github.com/j178/prek/pull/463 is the one that introduced this validation. Would be easy to ignore fields that start with `x-` there. I can send a PR too :)

udaya2899:
Following the convention used by tools like Docker Compose, OpenAPI, and GitLab CI, ignore any configuration keys starting with `x-` at any level (top-level config, repository entries, and hook entries). This allows users to define custom metadata and YAML anchors without triggering unexpected key warnings.

Fixes #2481

### DIFF
--- a/crates/prek/src/config.rs
+++ b/crates/prek/src/config.rs
@@ -1501,6 +1501,7 @@ where
     I: Iterator<Item = &'a str>,
 {
     for key in keys {
+        // Silently ignore extension keys starting with 'x-' (used for YAML anchors and custom metadata).
         if key.starts_with("x-") {
             continue;
         }
--- a/crates/prek/src/config.rs
+++ b/crates/prek/src/config.rs
@@ -1501,6 +1501,9 @@ where
     I: Iterator<Item = &'a str>,
 {
     for key in keys {
+        if key.starts_with("x-") {
+            continue;
+        }
         let path = if prefix.is_empty() {
             key.to_string()
         } else {
--- a/docs/reference/configuration.md
+++ b/docs/reference/configuration.md
@@ -47,6 +47,10 @@ The cooldown age is computed from the tag creation timestamp for annotated tags,
 
     The legacy `auto_update` key is still accepted as an alias for `update`.
 
+## Extension keys (`x-`)
+
+Any key starting with `x-` (a lowercase `x` followed by a hyphen) is silently ignored by `prek` at any level of the configuration (top-level, repository entries, and hook entries). This supports custom metadata and YAML anchors without triggering unexpected key warnings.
+
 ## Top-level keys
 
 ### `repos` (required)
--- a/docs/reference/configuration.md
+++ b/docs/reference/configuration.md
@@ -49,7 +49,7 @@ The cooldown age is computed from the tag creation timestamp for annotated tags,
 
 ## Extension keys (`x-`)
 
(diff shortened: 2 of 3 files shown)


udaya2899:
### Verification

  • Ran the snapshot review test suite (cargo test --test validate), confirming all tests pass.
  • Verified code formatting with `cargo fmt --check` and linting with `cargo clippy --workspace --all-targets -- -D` warnings (zero warnings).

docs/reference/configuration.md
```diff
@@ -47,6 +47,10 @@ The cooldown age is computed from the tag creation timestamp for annotated tags,
 
     The legacy `auto_update` key is still accepted as an alias for `update`.
 
+## Extension keys (`x-`)
+
+Any key starting with `x-` (a lowercase `x` followed by a hyphen) is silently ignored by `prek` at any level of the configuration (top-level, repository entries, and hook entries). This supports custom metadata and YAML anchors without triggering unexpected key warnings.
```
j178:
```suggestion
Any key starting with `x-` (a lowercase `x` followed by a hyphen) is silently ignored by `prek` at any level of the configuration. This supports custom metadata without triggering unexpected key warnings.
```

crates/prek/src/config.rs
```diff
@@ -1501,6 +1501,9 @@ where
     I: Iterator<Item = &'a str>,
 {
     for key in keys {
+        if key.starts_with("x-") {
```
j178, line 1505: Can you add a comment before this line?

udaya2899: Added a comment explaining why this is needed.

udaya2899:
CI failure seems to be a github flake: 

```
+[stderr]
+fatal: unable to access 'https://github.com/oven-sh/bun/': Failed to connect to github.com port 443 after 75010 ms: Couldn't connect to server
```

Can we retry?

j178:
Thanks!

udaya2899:
Really appreciate the quick turnaround. We'd like to use #2478 and this PR in our setup. Can I propose a new release with the next patch version `0.4.13`?

j178:
I'd like to wait until we've accumulated a few more changes before making the next release. This usually takes one or two weeks.

udaya2899:
Ah okay. That makes sense. Until then, it has to be built from source to use?

udaya2899:
https://github.com/udaya2899/prek/releases/tag/v0.4.13-dev I made a pre-release with `0.4.13-dev` on the fork I used to make the contribution. This works for now to share with our developers until an official 0.4.13 release comes out including this. Thanks again for your support. :)

PR #2486, opened 2026-08-05
Add `yaml-language-server:` comment to YAML sample config

Repiteo:
## Summary

Many IDEs feature automatic schema recognition of common YAML files, including `.pre-commit-config.yaml`. However, the schema loaded by default for these files is for base pre-commit, which this project has already diverged from. We already manually assign the schema used for TOML, but no equivalent exists for YAML. While no official schema specification syntax exists, the most commonly recognized one *by far* is `# yaml-language-server: $schema=...`.

This PR adds the above comment to the sample config for YAML files, making prek's schema IDE-aware regardless of which format is selected.


j178:
Thanks!

Issue #2505, opened 2026-08-07
Node hook install corrupts the user's git index when committing from a linked worktree


will-wright-eng:
### Before submitting

- [x] I have searched the issue list for similar bug reports.

### Summary

Running prek as the `pre-commit` hook during `git commit` in a **linked git worktree** (`git worktree add`) can replace the worktree's git index with the file tree of a cached hook repository. After the run, every index-reading git command fails:

```
fatal: unable to read d3e222c562aeccbbd7e19d83e81daf6c54c4410d
```

The staged blobs only exist in the hook repo's object database, not the user repo's. All hooks report "Passed" — the corruption is only discovered on the next index read (a later hook, or `git status` after the commit fails with `error: Error building trees`).

**Root cause chain:**

1. `git commit` exports `GIT_INDEX_FILE` to the hook. From a linked worktree these are **absolute** paths (`<main>/.git/worktrees/<name>/index`, plus absolute `GIT_DIR`); from a main worktree `GIT_INDEX_FILE` is relative (`.git/index`), which is why only linked worktrees are affected.
2. prek keeps these variables in its environment (its own git commands need them) and passes them to every subprocess.
3. Since #2394 (v0.4.11), node hook environments are installed by giving npm the cached clone as a `git+file://` URL. npm's pacote GitFetcher then shells out to git with the inherited environment.
4. `git clone` honors an inherited `GIT_INDEX_FILE`: its initial checkout writes the _user's_ index, filling it with the hook repo's tree.

The errant command, captured with `GIT_TRACE` during the repro below:

```
trace: built-in: git clone file:///…/prek/repos/29c5ebd8dba238ce /…/prek/cache/npm/_cacache/tmp/git-cloneXXXX --recurse-submodules
```

Triggering requires a **cold node env** (first install, or env invalidated by a prek upgrade). A warm repo cache does not protect — npm clones _from_ it. Python pre-commit strips exactly these variables for its internal git operations (`no_git_env()` in `pre_commit/git.py`); prek's own clone commands already do the same via `isolate_from_git_env()`, but the variables still leak through non-git subprocesses like npm.

**Minimal repro** (no real commit needed — simulates the env git exports to the hook):

```bash
tmp=$(mktemp -d) && cd "$tmp"
git init repro && cd repro
git config user.email t@t
git config user.name t

cat > .pre-commit-config.yaml <<'EOF'
repos:
  - repo: https://github.com/igorshubovych/markdownlint-cli
    rev: v0.45.0
[...]

Issue #2293, opened 2026-07-04
need to set git work tree


Konfekt:
### Summary

As discussed at  https://github.com/jdx/mise/discussions/10768, setting `GIT_WORK_TREE` in https://github.com/j178/prek/blob/ce4363d33e6ade4f3f8a46902193703204ad91f8/crates/prek/src/main.rs#L211 leads to side effects such as other tools (mise's pyenv)  overriding the repo, now set to the repo path. Since `pre-commit` does not seem to need it, one wonders whether `prek` neither? 

### Willing to submit a PR?

- [ ] Yes — I’m willing to open a PR to fix this.

### Platform

ubuntu 22

### Version

0.4.6

### .pre-commit-config.yaml

```yaml
    hooks:
      - id: verify
        name: mise verify (pre-push)
        entry: mise run verify
        language: system
        pass_filenames: false
        stages: [pre-push]
```

### Log file

```
2026-07-04T07:55:32.866387Z DEBUG prek: 0.4.5
2026-07-04T07:55:32.866441Z DEBUG Setting GIT_WORK_TREE to `/home/user/code/project`
2026-07-04T07:55:32.866463Z DEBUG Args: ["/home/user/.local/share/uv/tools/prek/bin/prek", "hook-impl", "--hook-dir", "/home/user/code/project/.git/hooks", "--script-version", "4", "--hook-type=pre-push", "--", "origin", "git@gitlab.dlr.de:fl-uas-iamosa-hap4/project.git"]
2026-07-04T07:55:32.870926Z TRACE get_root: close time.busy=4.16ms time.idle=2.66µs
2026-07-04T07:55:32.870951Z DEBUG Git root: /home/user/code/project
2026-07-04T07:55:32.870959Z DEBUG Found workspace root at `/home/user/code/project`
2026-07-04T07:55:32.870962Z DEBUG Found project root at ``
2026-07-04T07:55:32.870968Z DEBUG Loading project configuration path=.pre-commit-config.yaml
2026-07-04T07:55:32.871185Z TRACE Executing `/usr/lib/git-core/git cat-file -e 107178aaf102b8209e5eae9782e80cc49436de33`
2026-07-04T07:55:32.877597Z TRACE Executing `/usr/lib/git-core/git merge-base --is-ancestor 107178aaf102b8209e5eae9782e80cc49436de33 219ccbd57e068ffb91ddc1aafc6a15b166fdfbc3`
[...]

PR #2579, opened 2026-08-18
Expose the git work tree to hooks running outside the git root

jeromeheissler:
Committing from a linked worktree (`git worktree add`) makes Git export an
absolute `GIT_DIR` to hooks without a `GIT_WORK_TREE`. Git then treats the
current directory as the work tree root. Committing from a main worktree
exports neither, so only linked worktrees are affected.

prek runs each workspace project's hooks with the cwd set to that project's
directory. In a linked worktree that made a hook's own Git command resolve the
repository root to the sub-project directory: `git add -u` inside a
sub-project hook rewrote the whole index as if the sub-project were the
repository, staging every path outside it as deleted and giving the surviving
paths the sub-project's blobs.

prek already derives the correct work tree in `init_git_work_tree`, but #2356
scoped it to prek's own Git commands so hooks do not inherit an environment Git
did not expose. Keep that scoping, with one exception: hook subprocesses that
start outside the git root, which is exactly the sub-project case. Hooks
running at the git root are unaffected, so #2293 stays fixed.

Related: #2505 isolated the same environment leak for dependency installs, but
hook execution still inherits it.

### DIFF
--- a/crates/prek/src/git.rs
+++ b/crates/prek/src/git.rs
@@ -37,9 +37,10 @@ pub(crate) enum Error {
 pub(crate) static GIT: LazyLock<Result<PathBuf, which::Error>> =
     LazyLock::new(|| which::which("git"));
 
-// Git hooks can expose `GIT_DIR` without `GIT_WORK_TREE`. Keep the derived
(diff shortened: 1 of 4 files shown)


j178:
Thanks!

PR #2592, opened 2026-08-21
Preserve terminal theme contrast for status labels

### LINKED ISSUE #1310: Fix coloring of `Passed`
The `Passed` text is unreadable in certain terminal themes.
For example in [ghostty](https://ghostty.org) using the default background and foreground colors.

Image

gurgeous:
Love prek, I am using it in all my projects now. There is just one tiny thing I would like to fix... See #1310, which is still a  problem for many terminal themes (like mine).

We should probably be using fully specified ansi colors for bg/fg since you can't rely on the user's theme for bg colors. This PR switches to ANSI 256 colors (from ANSI 16). Also consider adding a space on either side of the text to make it prettier, but that's a separate thing. Thanks guys!

### NEW
new

### OLD
old

Fixes #1310

### DIFF
--- a/crates/prek/src/cli/run/mod.rs
+++ b/crates/prek/src/cli/run/mod.rs
@@ -1,3 +1,5 @@
+use owo_colors::{Style, Styled};
+
 pub(crate) use filter::{
     CollectOptions, FileSelection, FileTagCache, FileTagFilter, HookFileFilter, ProjectFiles,
     RunFileIndex, RunInput, collect_run_input,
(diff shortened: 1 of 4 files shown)


pygarap:
@gurgeous Why are you using ANSI 256 colors instead of explicit 24-bit Truecolor?

gurgeous:
> @gurgeous Why are you using ANSI 256 colors instead of explicit 24-bit Truecolor?

I have no preference personally. There are still a few terminals that don't support truecolor. Last time I looked the two major laggards were gitlab CI and Apple Terminal pre-Tahoe. Popular software (like prek) might encounter those in the wild, so sometimes it's easier to just use ansi 256. Totally up to you, of course.

j178:
@gurgeous @pygarap Hi, I tried a different approach that preserves the terminal theme’s contrast. Could you check whether this improves readability in your terminal?

gurgeous:
This is an improvement in most cases. Still pretty easy to find themes that look bad. For example, ghostty ciapre. This is just how it goes when you try to reverse the user's completely unknown color theme...

image

j178:
Thanks for confirming! I think any terminal theme should at least keep these predefined colors readable against its default background, so reverse video seems like the safest option here.

gurgeous:
BTW I added a tool on ansi.md to illustrate for next time, in case anyone is interested. Still love prek and use it in all my projects, thank you for this fix. See https://ansi.md/bad-colors.

"theme bg" on "ansi green" looks ok, though there are a few themes that are hard to read.

PR #2613, opened 2026-08-25
Add configurable toolchain source preferences

### LINKED ISSUE #2612: Add an option to always prefer managed toolchain
### Discussed in https://github.com/j178/prek/discussions/2607

<div type='discussions-op-text'>

<sup>Originally posted by **libvoid** August 25, 2026</sup>
Hello,

As of now, `prek` always tries to use the system toolchain first.

> `language_version` is treated as a version request, not a single pinned value. For languages that use semver requests, you can specify ranges (for example `^1.2`, `>=1.5`, `<2.0`).
>
>**Special values**:
>    * default: use the language’s default resolution logic.
>    * system: do not download a new toolchain. A compatible toolchain already managed by prek may still be reused, and project metadata can further constrain the required version.
>
>**Language-specific behavior:**
>
>    * Python: passed to the Python resolver (for example `python3` or `python3.12`). May trigger toolchain download.
>    * Node: passed to the Node resolver (for example `20`, `18.19.0`). May trigger toolchain download.
>    * Go: uses Go version strings such as `1.22.1` (downloaded if missing).
>    * Rust: supports rustup toolchains such as stable, beta, nightly, or versioned toolchains.
>    * Other languages: parsed as a semver request and matched against the installed toolchain version.
>
> https://prek.j178.dev/languages/?h=language#toolchain-management-and-language_version

This makes sense to do this by default but I think it would be useful to allow the user to force always using a prek-managed toolchain.

Here is an exemple: I have ruby installed on my machine but without `ruby-dev`, installing some gems will fail if they depend on prism (see #2605) because it is not installed. I would prefer in this specific case to no rely on the distribution `ruby-dev` package and can't uninstall `ruby` to force `prefk` to install the toolchain.

**Note :** the current workaround is to set for a langage version not available on the system.</div>

j178:
Closes #2612 
Closes #2607

### DIFF
--- a/crates/prek/src/languages/version.rs
+++ b/crates/prek/src/languages/version.rs
@@ -1,4 +1,6 @@
+use std::ffi::OsStr;
 use std::fmt::{self, Display};
+use std::path::{Path, PathBuf};
 use std::str::FromStr;
 
 use crate::config::{Language, LanguageVersion, ToolchainPreference};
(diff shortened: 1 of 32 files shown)


crates/prek/src/languages/version.rs
j178: Can be removed

PR #2670, opened 2026-09-01
Revert unsupported YAML tag rejection in check-yaml

j178:
Reverts #2656 

The `reject_unsupported_tags` behavior rejects valid application-specific YAML tags used by `mkdocs.yml`, including `!!python/object/apply:pymdownx.slugs.slugify`. Restore the previous `check-yaml` behavior and parser versions.


j178:
Reverted this revert in #2672

PR #2682, opened 2026-09-03
Add a `check-jsonc` builtin hook

### LINKED ISSUE #2383: Feature Request: Support JSON files with comments
I have a few JSON files that contain comments. Usually, I use json5 but there is also `.jsonc` which seems to have wider support (it does not support trailing commas though). For example, I am attempting to use a config file for Renovate Bot as `renovate.json` (could be `renovate.jsonc` if that's easier) which includes comments. Another one is `.vscode/settings.json`.

It would be great if the JSON hooks could support files with comments.

clbarnes:
## Summary

Resolves #2383. See [this discussion](https://github.com/j178/prek/issues/2383#issuecomment-5524044982) for the distinction from `check-json5`.

Adds a builtin `check-jsonc` hook for `.jsonc` files, using the [`jsonc-parser` crate](https://crates.io/crates/jsonc-parser). Trailing commas are rejected unless `--allow-trailing-commas` is set. Duplicate keys are rejected.


clbarnes:
The remaining test failure seems to be a failed attempt to install golang, which I can't claim responsibility for ;)

clbarnes:
@j178 thanks for taking a look! However, I don't agree that trailing commas should always be allowed, for the same reason that JSONC validators shouldn't just use the JSON5 hook. Some JSONC parsers do not allow trailing commas, and this is [compliant with the spec](https://jsonc.org/#appendix-a-trailing-commas-and-jsonc), and common JSONC use cases like VSCode warn against the use of trailing commas. We shouldn't presuppose which JSONC implementation or configuration the prek user is planning to validate for, so we should allow them to either allow or disallow trailing commas. Any other behaviour is not a faithful representation of the JSONC spec.

j178:
Agreed, I've added `--allow-trailing-commas` back.

PR #2683, opened 2026-09-04
chore: update asyncband to 0.7

tisonkun:
## Summary

Update AsyncBand to 0.7 and enable only the `once-cell`, `once-map`, and `semaphore` primitives used by prek.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -222,12 +222,11 @@ dependencies = [
 
 [[package]]
 name = "asyncband"
-version = "0.6.7"
+version = "0.7.0"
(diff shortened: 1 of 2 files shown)


j178:
Thanks!

tisonkun:
@j178 Ah.

Due to some compliance issue the 0.7.0 would be yanked and I'll cut a 0.7.1 release later. There is no bug anyway, but some ASF release policy issues should be fixed.

tisonkun:
I'll prepare a follow-up PR and feel free to ping me if anything of asyncband blocks your workflow.

PR #2684, opened 2026-09-04
chore: update asyncband to 0.7.1

tisonkun:
## Summary

Update AsyncBand from 0.7.0 to 0.7.1 because 0.7.0 has been yanked.


j178:
Thanks for the follow-up!

